PR #5, opened 2018-04-03
Problem: There is no catalog for discovering assets

diminator:
Solution: SONAR OEP initialization


mikera:
Can you describe the problem more, perhaps with a definition of what you expect from an asset catalogue and some examples of usage?

3/README.md
```diff
@@ -0,0 +1,45 @@
...
+name: Searchable OceaN Asset Registry 
+type: Standard
+status: Draft
+editor: Dimitri De Jonghe <dimi@oceanprotocol.com>
+contributors: Aitor Argomaniz <aitor@oceanprotocol.com>, Mike Anderson <mike.anderson@dex.sg>
+```
+
+## Abstract
+
+The SONAR OEP is any software, protocol or interaction that allows actors
```
mikera: I assume we aren't talking about "any" here but either a specific reference implementation or an API that such software can use?

aaitor:
I had a problem with the Markdown editor, I will restore the previous version of the README.md

3/README.md
```diff
@@ -0,0 +1,174 @@
+```
+shortname: 3/SONAR
+name: Searchable OceaN Asset Registry 
+type: Standard
+status: Draft
```
diminator: Raw

3/README.md
```diff
@@ -0,0 +1,174 @@
...
+The data services require to be discoverable and queryable, as well as consumable.
+
+## Specification
+
+The **Assets** information should be managed using an API. This API should exposes the following capabilities:
+
+* Adding an Asset
+* Get an Asset
+* Update an Asset
+* Disabling an Asset
```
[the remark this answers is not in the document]
aaitor: Disabling implements a "soft delete". Because it's not possible to delete from an immutable data-store, the only way to "delete" contents is to provide an attribute or flag specifying if the content is active or not. This method implement the logic to disable an existing Asset. It could be useful when you registered an Asset by error or you want to unpublish an  Asset.

3/README.md
```diff
@@ -0,0 +1,174 @@
...
+* Adding a Provider to an Asset
```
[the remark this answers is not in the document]
aaitor: The provider is the one giving access to the Asset. For example a provider could be a guy/company running IPFS nodes to store/serve some Ocean datasets.
The Provider will get tokens in reward and must provide a Proof of Service.

3/README.md
```diff
@@ -0,0 +1,174 @@
+```
+shortname: 3/SONAR
+name: Searchable OceaN Asset Registry 
+type: Standard
+status: Draft
+editor: Dimitri De Jonghe <dimi@oceanprotocol.com>
```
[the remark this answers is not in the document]
aaitor: change applied

3/README.md
```diff
@@ -0,0 +1,174 @@
...
+contributors: Aitor Argomaniz <aitor@oceanprotocol.com>, Mike Anderson <mike.anderson@dex.sg>, Samer Sallam <samer@oceanprotocol.com>, Kiran Karkera <kiran.karkera@dex.sg>
```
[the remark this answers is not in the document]
aaitor: change applied

3/README.md
```diff
@@ -0,0 +1,174 @@
...
+
+This OEP does not focus on metadata structure, staking or curation. 
+It's purely the mechanics of publishing and retrieval of assets.
+Optionally this might contain a basic framework for the actor registry without the identity platform. 
+
+## Motivation
+
+Ocean network aims to power marketplaces for relevant AI-related data services. 
+The data services require to be discoverable and queryable, as well as consumable.
+
```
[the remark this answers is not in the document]
aaitor: refactored adding this description and taking into account the PROVIDERS

3/README.md
```diff
@@ -0,0 +1,183 @@
...
+ASSETS are DATA objects describing RESOURCES & SERVICES under control of a PUBLISHER.
+A PUBLISHER is incentivized to PUBLISH ASSETS in order to make them discoverable for third parties.
+A PROVIDER can discover ASSETS and give access to those ASSETS getting tokens in reward.
+
+ASSETS metadata are STORED on a decentralized REGISTRY.
+ASSETS metadata can be UPDATED. Optionally with versioning.
+ASSETS metadata can be RETRIEVED from the REGISTRY.
+ASSETS content can be RETRIEVED from the PROVIDER.
```
diminator: Is this a responsibility for the ASSET REGISTRY?
Could be a responsibility for they access control layer

3/README.md
```diff
@@ -0,0 +1,183 @@
...
+This OEP does not focus on metadata structure, staking or curation. 
+It's purely the mechanics of publishing and retrieval of assets.
+Optionally this might contain a basic framework for the actor registry without the identity platform. 
+
+## Motivation
+
+Ocean network aims to power marketplaces for relevant AI-related data services. 
+The data services require to be discoverable and queryable, as well as consumable.
+
+ASSETS are DATA objects describing RESOURCES & SERVICES under control of a PUBLISHER.
```
diminator: ASSETS or ASSET metadata?

diminator: Isn't the PROVIDER controlling the SERVICES rather than PUBLISHER?

3/README.md
```diff
@@ -0,0 +1,183 @@
...
+A PUBLISHER is incentivized to PUBLISH ASSETS in order to make them discoverable for third parties.
+A PROVIDER can discover ASSETS and give access to those ASSETS getting tokens in reward.
```
diminator: I'm struggling a bit with the fact that a PUBLISHER can publish assets on behalf of the PROVIDER
Feels like it's complicating things.

Would the responsibility of rights delegation of the ASSET not be outside of this layer?
For IP-related or hosted ASSETS, this might be important as we might include group ownership, referrals, consignees...

3/README.md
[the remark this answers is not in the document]
diminator: should we link to a glossary for ocean-wide definitions?

3/README.md
diminator: I think the main attack vector is when a PUBLISHER would be able to misguide consumers with ASSETs that are not under his (delegated) control

3/README.md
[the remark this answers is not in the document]
mikera: We could call it "Delete" if that makes things clearer to users. The fact that it is implemented in the underlying storage as a soft delete is an implementation detail that we should try to avoid exposing in a public API.

3/README.md
mikera: I agree, the provider controls the services. Publisher is the creator / owner / custodian of the IP embodied in the asset. Many providers may serve up the same asset (indeed, I think this is essential if the curation markets are going to work!).

3/README.md
mikera: If the scope here is purely the registry of assets, then provider-related stuff (i.e. actually uploading and downloading the asset data) can be moved to a separate OEP. I think this is a good idea, keep this one focused on the registration of asset metadata itself.

3/README.md
mikera: Should be asset metadata I think. Provider controls services, publisher controls asset metadata (and is responsible for the initial upload of the asset itself to one or more providers able to offer a storage service)

3/README.md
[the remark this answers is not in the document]
shark8me: I noticed that other (blockchain) APIs use the term "retire", I think it fits better

3/README.md
shark8me: I'd prefer we say "Assets are data objects owned by a publisher". "under the control of" could be misleading, as Providers own the role of 'control'

3/README.md
shark8me: I believe the confusion (over naming) arises because we aren't using the conventional meaning of the term "publisher". If we take the literature analogy, the AUTHOR is the owner, and the PUBLISHER is one who makes the asset available. Its the same with the [music industry](https://en.wikipedia.org/wiki/Music_publisher_(popular_music).   

If we use OWNER instead of PUBLISHER, the confusion could be avoided by a large extent.

3/README.md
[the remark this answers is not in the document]
shark8me: I find the API hews too close to the "CRUD" model (for DB design), and the **verbs are not intuitive**. When I compare other dataset APIs such as [mapbox](https://www.mapbox.com/api-documentation/#datasets), [Bigquery](https://cloud.google.com/bigquery/docs/datasets#bigquery-create-dataset-java), 

The *verbs* common across these APIs are:

* Create asset/dataset (metadata such as name, column info etc)
* Add/Modify rows or features (add rows to a dataset)
* List
* Retrieve / Download (the data, not properties)
* Update Properties (not data/rows)
* Delete.

To me, "add asset", "get asset", "update asset", could mean either asset metadata or asset data. Anyone reading the API method names should **not** have this confusion.

3/README.md
shark8me: Maybe a naive question.

Why not just rename **asset** with **dataset**? The term is universally understood, all the commonly used libraries (Tensorflow/Spark/Bigquery/Azure), use dataset and not asset. 
That way we don't have to resort to recursive definitions such as "Assets are data objects", where the reader then asks, what is a "data object".

Stating that a "Dataset D is Owned/authored by user1 and can be retrieved from Provider P", is clear without having to first clarify what is a data asset/publisher/provider.

3/README.md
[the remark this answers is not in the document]
aaitor: About the disable/delete term, the original name used was delete. It was changed to disable because a previous recommendation. I'm OK to change it again, but I think it's not a big deal.

3/README.md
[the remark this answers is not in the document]
aaitor: :+1:

3/README.md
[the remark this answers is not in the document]
aaitor: Verbs used are the HTTP standard: http://www.restapitutorial.com/lessons/httpmethods.html
In REST, primary data representation is a Resource, that it's different to HTTP RPC where typically is a Verb or Action:
https://www.ics.uci.edu/~fielding/pubs/dissertation/rest_arch_style.htm#sec_5_2_1_1

3/README.md
aaitor: :+1: 
Because the main "Provider" functions (upload/download binary) are out of the keeper scope, I would have it too in a different OEP.

3/README.md
aaitor: I think term asset makes sense when we want to abstract a dataset and a service using a common term. Meaning of asset is: "a useful or valuable thing or person"
So if we want to be more specific about dataset's, services and their metadata we could expose those in this way:
Get Asset Metadata: **/api/v1/keeper/metadata/asset/{assetId}**
Get Service Metadata: **/api/v1/keeper/metadata/service/{serviceId}**

shark8me: In a discussion with Mike, I understand that **Asset** was used because the network consumers could upload/download algorithms and other computation implementations, in addition to fixed or immutable datasets. In that case, assets is fine as an overarching term.

3/README.md
shark8me: The Glossary does contain the term Data Owner and Publisher. The description in this file should mirror what the Glossary says (right now, it doesn't. It should include the (role) description of data owners, publishers, and the probable area of overlap).

3/README.md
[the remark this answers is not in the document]
shark8me: Your point (about REST API design) is certainly valid. My concern was about a higher level concern that the API methods should be meaningful, without having to second guess whether the data or metadata was being created or deleted. 

We should "separate concerns" about what the API will allow, vs what it is implemented in (e.g. REST).  
We can certainly have an implementation where ** Create Asset** is called on "/api/asset/metadata" and **Create Rows** is called on "/api/asset/data". 

I propose that the methods:
* Create Asset
* Update Asset
* Retire Asset

refers to the metadata (as discussed above), and have other methods for the **data** , such as:

* Add/Modify/Delete Asset **Data** (if the asset is not immutable)

and for immutable assets:

* Upload Asset Data 
* Retrieve Asset Data. 

Is that an agreeable option?

3/README.md
aaitor: :+1:

3/README.md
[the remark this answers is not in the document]
aaitor: This is quite similar the API we defined:
* Create Asset - ASE.001
* Update Asset - ASE.003
* Retire Asset - ASE.004

As we discussed during the London sessions, Assets are immutable in the way an update of the Asset (ie. content of the dataset) requires to create a new Asset (ASE.001).
* Upload Asset Data (ASP.001)
* Retrieve Asset Data (ASP.002)

I hope this helped to clarify

3/README.md
diminator: So, what about this:

ASSET metadata are DATA objects describing RESOURCES & SERVICES.

I think this OEP should be limited to only PUBLISHER and not PROVIDER.
Otherwise we'll end up including the licensing schema here, whilst this should be storage and discovery of ASSET METADATA.

3/README.md
diminator: What about this?

A SONAR client can discover ASSETS by querying the ASSETS registry.

3/README.md
```diff
@@ -0,0 +1,186 @@
+```
+shortname: 3/SONAR
```
mikera: Wrong OEP number?

3/README.md
```diff
@@ -0,0 +1,186 @@
...
+## Abstract
+
+The SONAR OEP is a specification for Ocean Protocol to publish and discover ASSETS such as data sets, algorithms, infrastructure and so on. ASSETS are described by structured METADATA.
+
+Optionally this OEP extends to hold other datatypes such as actors.
+
+## What not
+
+This OEP does not focus on the storage backend, METADATA structure or incentives around signalling. 
+It's purely the interface for publishing and discovery of ASSETS.
```
mikera: Isn't that meant to be OEP 14? Is this OEP redundant now?

aaitor:
OEP out of date, will be consolidated by OEP-14

PR #29, opened 2018-04-19
OEP3/4/5 Architectures

diminator:
- [x] OEP3: network architecture
- [ ] OEP4: keeper architecture
- [x] OEP5: client architecture


3/README.md
```diff
@@ -0,0 +1,154 @@
...
+
+In the remainder of this document we refer to:
+- AGENT: client-side software responsible for constructing transactions.
+Examples are CONSUMERS, PROVIDERS, MARKETPLACES, PUBLISHERS and so on.
+- KEEPER: consensus-side software responsible for validating transactions, 
+smart contracts execution, consensus algorithm coordination and decentralized storage.
+
+
+![Contract Centric Architecture](images/contract-centric-arch.png)
+
```
shark8me: The network architecture should also include Publishers, who (in the role of asset owners), should have the right to accept/deny/change the price of assets. In some scenarios, the consumption of an asset is a 4-way tango with the consumer, marketplace, provider and publishers. That's not apparent in the diagram.

3/README.md
```diff
@@ -0,0 +1,154 @@
...
+A work-dispatch engine with the following capabilities:
+
+* Interacts with the clients through transactions.
+* Performs compute intensive jobs such as mining and proofs validation
+* Can challenge clients to provide succinct proofs
+* Is fault tolerant
+
+
+### Ocean Agent
```
shark8me: "Ocean Agent": The term "agent" is often conflated to mean "autonomous agent" or " computer program that acts for a user or other program" (from wikipedia.) What is described, is it implements an API that enables to consumer to talk to an Ocean Node. 
Why not just call it Ocean API? 
If indeed we call it an agent, on whose behalf does it act?

4/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+## Responsibilities
+
+The main responsibilities of the Keeper are:
+
+* Expose an interface to receive requests requiring to execute some actions. Initially this will expose a HTTP RPC interface
+* Validate the input requests implementing the RBAC system 
+* Persist the Metadata and Transactions information On-Chain
+* Maintain the decentralized state including the Proofs of Service, Contracts bytecode and Actors registry
+* Implement the Curation Market Business logic for data and services
```
shark8me: Uniform naming: prefer **assets** instead of ** data**.

4/README.md
```diff
@@ -0,0 +1,185 @@
...
+* Manage and track the IP ownership of assets and services during their creation and access
+* Provide the Token curated registry of actors
+* Maintain the P2P consensus between the Keeper nodes
+* Orchestrate the Keeper business logic interacting with the different external resources
+* Manage the external Processing entities providing computing capabilities
+* Manage the external Storage entities providing storage capabilities
+
+## Decentralized State
+
+![Decentralized State](images/decentralized-state.png)
```
shark8me: It is not clear if this diagram intends to depict a stack (like a tcp/ip stack), or a set of components used in the decentralized state.

4/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+
+**Ocean Core** itself does not store the content of the Assets, instead, it links to data that is stored, and provides mechanisms for access control. At the same time Keepers store a different kind of metadata:
```
shark8me: What is **Ocean Core** mean here (this is the first time its used in this document, if it is a component, define what is meant by it).

4/README.md
aaitor: :+1:

3/README.md
aaitor: I think responsibilities of Agent are more than exposing the API. Some of the things to do are related with events listening, orchestration, etc.

4/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+* Verification - Allowing to handle the actors and assets verification and whitelisting process.  This information should be managed using the Decentralized VM and Ocean DB. Should provide the following functions:
+  - Definition of Verification challenge - A pre-appointed verifier and/or the client challenges the service to provide proofs that the requested service is delivered according to integrity specifications.
+  - Provide a Verification Proof - The service accepts the challenge, computes the proof and stores this on-chain with a reference to the contract.
+  - Get details of a Verification Proof - Retrieval of Verification Proof information.
+  - Actor Whitelisting - A verifier send a whitelisting request approving or denying to an actor
+  
+* Wallet - Allowing the manage the basics about the user wallets.  This information should be managed using the Decentralized VM. Should provide the following functions:
```
shark8me: What **User** refers to is unclear. Is it instead, all **actors** in the network?

5/README.md
```diff
@@ -0,0 +1,245 @@
...
+The Agent also will orchestrate the interaction with the Keeper components, allowing to provide a high level view from the consumer side, interacting with the decentralized VM and the Ocean DB.
+
+### Interfaces
+
+In charge of receiving the external requests to interact with the system. Initially the Ocean Agent will expose a HTTP RESTful API, but is designed to expose the API’s or consume requests in different ways.
+Because of that the Ocean Agent should expose the API’s in different formats allowing integration mechanisms that can be used depending on the use case. The initial consumption mechanisms could be:
+ 
+* **Request/Response** - Provided by the RESTful and the RPC interfaces. Those allow a request/response integration. The API will expose different HTTP methods implementing the defined actions.
+
+![Request/Response](images/req-response.png)
```
shark8me: This image (and the ones similar to this) should also include providers

5/README.md
```diff
@@ -0,0 +1,245 @@
...
+
+* **Async Websocket** - Provided by the Websocket interface. Useful when some actors need to be subscribed to the changes happening in the database. For example, if a change in a contract is happening.
```
shark8me: The term "websocket" is a specific type of  async communications. Since most, if not all actors are servers (and not browsers), **webhooks** should be the method used (see [this description](https://stackoverflow.com/questions/23172760/differences-between-webhook-and-websocket) of the difference between websockets and webhooks). 

Given that pub/sub/event driven interface is mentioned next, I don't see a need to describing a separate async interface.

5/README.md
```diff
@@ -0,0 +1,245 @@
...
+
+Those capabilities will be integrated from the Ocean Agent using different protocols. 
+
+The Keeper interface module should implement an extensible interfaces system allowing to plug different communication protocols to establish the communication between the Ocean Agent and the the Keepers network. This component is in charge of the following capabilities:
+
+* Interact with the Keeper components (Decentralized VM, Ocean DB)
+* Compose the transactions payload necessary by the Keeper nodes
+* Orchestrate the execution of multiple Keeper requests when be necessary
+
+Initially, HTTP RPC is the easiest candidate to integrate in the communication with the Decentralized VM. A part of HTTP RPC, in following iterations this component could interface with HTTP RESTful API’s and using the Interledger Protocol (ILP).
```
shark8me: This section is needs to be edited for clarity.

It makes sense to say that components A and B are integrated. HTTP RPC/REST are communication protocols, and not components. 

Instead, it could be 
"Initially, communication with the Decentralized VM is through a HTTP RPC interface. Subsequent iterations will add support for HTTP REST as well as ILP".

diminator:
> In the figure of building-blocks.png, the "worker" module is in charge of computing inside Keeper such as validate the transaction. Please correct me if I'm wrong. In my understanding, they cannot perform the compute-intensive task such as AI algorithms, which shall be handled by off-chain computing with truebit or enigma.

- The worker is meant to be a "mining-like" function that secures the network. Mining farms in BTC/ETH for example have HPC capabilities. The worker aims to unlock the block reward from the network - this could be classical PoW but also fishing for invalid proofs, bridges, ...
- the transaction validation for the VM happen in the VM node
- the validation for the decentralized DB happen in the DB node

5/README.md
```diff
@@ -0,0 +1,245 @@
...
+     
+* Direct messaging between parties - Allowing for example the agree Contracts terms before to formalize the contract or direct sharing of Assets consumptions information between the provider and the consumer. 
+* Assets transferring between different Ocean actors - In the actors that are giving access directly to some assets without using a third-party provider, would be possible to share directly the Assets between parties.
+     
+The implementation of the P2P communication is highly related with the existing p2p libraries. 
+
+
+#### Interfaces with external providers
+
+Ocean Agent should provide a pluggable mechanism allowing to interact with external providers. It could be:
```
shark8me: This section is quite vague. 
In computing terminology, a plugin **is a software component that adds a specific feature to an existing computer program** [link](https://en.wikipedia.org/wiki/Plug-in_(computing). 

If the intention was to state that Ocean components can interface with other computing services or providers, with interface such as REST/RPC , then the use of the term plugin is misleading IMO.

5/README.md
```diff
@@ -0,0 +1,245 @@
...
+* **Storage or Data Providers** - In charge of providing storage services off-chain or on-chain. Like Amazon S3, IPFS, etc.
+* **UX/UI Providers** - In charge of providing visual interfacing with the system.
+
+Because is difficult to define upfront the different providers to integrate, it’s important to implement a pluggable mechanism allowing to extend the systems supported by the system.
+
+![Ocean Agent Plugins System](images/agent-plugins.png)
+
+In the above picture the Storage, Computing and UX/UI interfaces have the responsibility of modelling the interaction with the external systems. Having this approach, support additional providers would require only the implementation of the communication with the new provider, but the integration with the rest of the system should be simpler.
+
+Depending of the implementation of the system, the usage of one plugin or another one, could be made by configuration or using dependency injection.
```
shark8me: It is unclear why Java or Object oriented programs are mentioned here. Is it an example? If the plugin system runs in a JVM or OSGi container, then its apt.

5/README.md
```diff
@@ -0,0 +1,245 @@
...
+    ..
+}
+```
+
+## Access Control
+
+![Agent Access Control](images/agent-access-control.png)
+
+
+Access Control system implements the architecture where external users or applications are Authenticated and Authorized in the system, allowing (or denying) the management of the resources.
```
shark8me: what's the difference between an **external** and an **internal** user? Are these just actors?

5/README.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Privacy Management
+
+TODO: Protocol to negotiate privacy requirements
+
+## Events Watcher
+
+An important part of the Smart Contracts implementation in the Decentralized VM is the triggering of Events. Those events could expose system notifications when some relevant actions are happening (ie. a new Asset is registered, an Asset is curated, etc.).
+
+From the Ocean Agent side, the Events Watcher component will be in charge of watching those events to trigger some further actions.
```
shark8me: Is the Event watcher part of the decentralized VM, or the Keeper?

diminator: Ocean Agent

diminator:
> I feel confused about the architecture in the figure: "decentralized database" is a third-party service (e.g., BigChainDB) since it is not a part of Ocean Node itself. From this point of view, this is a top-level architecture, therefore, it shall include "decentralized computing" providers rather than "workers". On the other hand, if this figure shows the structure of Ocean Node, it shall not include Database for the same reason. Maybe I have some mistake but just bring up this confusion

The decentralized DB and workers are not refering to services from PROVIDERS. They are keeping functions: ie make sure the shared stated is kept

3/README.md
shark8me: I'm ok with that, I was just pointing out that the name "Agent" does not suit the responsibilities, perhaps a different name is better, since "Agent" connotes that it acts as a proxy (for some other entity)

3/README.md
diminator: I think we can have multiple roles in the system, and this picture is only depicting a few.
From a protocol and network point of view these roles are dynamic and depending on the contract type or off-chain agreements. Hence I classified them as a form of AGENTS.

I think that if we consider these roles hard coded right now, we won't be able to distill a clean architecture

diminator:
> I have one fundamental issue that I feel must be addressed before we merge or move to a DRAFT status on these OEPs. It feels like the "Ocean Agent" is becoming more like a heavyweight enterprise middleware system, responsible for routing and orchestrating everything. I think it should be more of a lightweight agent mainly responsible for exposing the Ocean Protocol API and interfacing with the decentralised network(s) and services.

Agreed that it's heavyweight - consider this a description for a full node implementation. My suggestion would be to keep this full reference as is and soon propose a lightweight implementation (maybe based on interledger/STREAM or the likes).

> A) People running agents will need to do a lot of extensive configuration and customisation (e.g. integrating to an implementation of an Ocean DB, pluggin into custom ACL systems). This will significantly raise the bar to adoption, and will create problems when configuration requirements conflict (e.g. what if you want to run multiple Ocean DBs? How do I route to multiple service provider instances?). It would be better if most participants could just run a standard Agent with minimal configuration.

We're assuming that might be the case in the beginning - We could standardize configuration templates for known services and distinguish various types of agents (like a super-lightweight agent)



> B) By assuming everything going through AGENT-AGENT comms we seem to preclude the possibility of lightweight participants who don't need to run a network-accessible agent. I think AGENT-SERVICE comms may make more sense in many cases (e.g. a marketplace that opens an API to provide asset metadata subject to domain-specific ACL restrictions). You also typically cut out a whole level of indirection doing this, i.e. the comms can be USER-AGENT-SERVICE rather than USER-AGENT-AGENT-SERVICE

yes, definitely probable - as stated above, I think this belongs to a light client OEP

> C) The plugin system looks like it could quickly become over-complicated. Different types of services will require different interfaces, and all of this will start impacting Ocean Agent code, i.e. we run the risk that application level logic will start seeping into the Agent which should really be about protocol/network level functionality. The caller should understand what type of service is being consumed and how to do so (possibly by inspecting metadata) but I don't see why the Agent needs to care about this.

Indeed, no application logic should leak. The plugin only cares about translating the connection details. Hence, the plugin needs to be able to communicate with the service integrity layer:
- provide proof of service
- access control and/or privacy



> D) All the above issues will be exacerbated by the problem of different people running different agent versions in a decentralised setting. Enterprise middleware is often changing as requirements evolve, we need the Ocean Agent to be stable if we arte going to govern the network effectively.

The setup of the AGENT is modular and layered - this means that layers can be omitted and simple modules can be used instead of complex ones. This is the crux of the design: allowing a lot of flexibility.
If that wasnt clear from the design then we need to iterate a bit more on the specification:
- explain layering and modularity
- show examples of full AGENT vs lightweight AGENT running a similar stack with simplified blocks

PR #30, opened 2018-04-24
OEP 13 - Actors registry 

aaitor:
…block


aaitor:
I was thinking about that and I see two possibilities:

- As you mention, in the actor registering process the balance is initialized in the Decentralized VM. Usually, this balance is 0 until an initial transfer of funds happens. Initialize the balance in the Decentralize VM has the cost of writing on-chain.
- As an alternative option, I think maybe it's not necessary to write the user balance on-chain until some tokens are transferred to the account for the first time. In this case, the registering only create the user metadata in Ocean DB. The cost of registering an Actor would be zero. And only the users setting up some balance will create this row on-chain, paying the blockchain fee for this writting. This scenario would allow users registering without any fee, that means a lower barrier of entry for new users.

Does it makes sense you @gongf05 ?

shark8me:
I have a question about user/actor information..

If a data publisher wishes to publish a dataset anonymously, what are the options available to him/her?

* Can the user withhold information such as number of tokens/name from being visible on the (public?) chain?
* Can the user information be made available, on request, to a select whitelist of actors?

13/README.md
```diff
@@ -0,0 +1,628 @@
...
+
+The above diagram shows the high level interactions between the components involved:
+
+* The ACTOR interacting with the Actors Registry will send an API request to the AGENT
+* The AGENT MUST validate the basic parameters sent by the ACTOR
+* The AGENT MUST authenticate the ACTOR sending the request (but in the new actor method)
+* The AGENT MUST orchestrate the ACTOR authorization using the KEEPER
+* The AGENT MUST handle the account management (creation,unlock,etc.)
+* The DECENTRALIZED VM store the ACTOR essential information (public key)
+* If OCEAN DB is enabled by the user, will store the ACTOR metadata
```
mikera: The network needs to store metadata about the Actor regardless of whether an Ocean DB is configured locally.  Presumably, if the metadata is to be stored off-chain, then an on-chain link needs to be provided to the service provider / agent that acts as an authoritative metadata source? 

OEP needs to address this issue.

13/README.md
```diff
@@ -0,0 +1,628 @@
...
+        }]
+	}
+}
+```
+
+#### Accounts Management
+
+The AGENT node will be in charge of manage the Accounts creation. To do that the AGENT will interact with the **Accounts API** provided by the **Parity** client.
+
+The API method to integrate is the [personal::newaccount](https://wiki.parity.io/JSONRPC-personal-module#personal_newaccount).
```
mikera: Seems like this is an implementation detail. Should not be in the overall OEP for actors registration.

13/README.md
```diff
@@ -0,0 +1,628 @@
...
+|:----------|:-----|:------------|
+|actorId    |address|Owner of the Asset|
+|state      |uint  |TCR state of the user|
+  
+Using any of the existing web3 implementation library (web3.js, web3.py, web3.j, etc), it's possible to interact with the VM Smart Contracts.
+
+
+#### Interaction with Ocean DB
+
+The integration with OCEAN DB is optional, so only will works if an implementation backend is provided.
```
mikera: This section doesn't make sense. The protocol needs to work regardless of whether people have configured a DB. 

Different actors may have agents configured in different ways so it is unclear how this is supposed to work in the case that some actors have DBs, some don't, and some have different DBs. This appears to be a fundamental design flaw in the Agent architecture, see my separate comments on OEP 3/4/5

13/README.md
```diff
@@ -0,0 +1,628 @@
...
+```json
+{
+    "actorId": "0x8f0227d45853a50eefd48dd4fec25d5b3fd2295e",
+    "state": "DISABLED"
+}
+```
+
+#### Accounts Management
+
+The Accounts Manager components it's not involved in this method.
```
mikera: Seems redundant to include this section, in that case?

13/README.md
```diff
@@ -0,0 +1,628 @@
...
+    "actorId": "0x8f0227d45853a50eefd48dd4fec25d5b3fd2295e",
+    "state": "DISABLED"
+}
+```
+
+---
+<a name="asset-tcr"></a>
+### Token Curation Registry (TCR) of Actors 
+
+Token curation registry (TCR) is used to maintain a list of virtuous actors through challenge-voting process:
```
mikera: Is TCR mandatory for registering an actor? Why do we need it if so? 

If not I suggest moving to a separate OEP.

13/README.md
```diff
@@ -0,0 +1,628 @@
...
+The KEEPER::Decentralized VM will store the essential user information to allow the implementation of the Actors TCR.
+It means the system MUST NOT store any personal information, enabling PRIVACY and ANONYMITY.
+
+The **KEEPER::Decentralized VM** will persist the following information:
+
+| Attribute | Type | Description |
+|:----------|:-----|:------------|
+|actorId    |address|Owner of the Asset|
+|state      |uint  |TCR state of the user|
+
```
mikera: I think this needs an optional marketPlaceID in the case that either:
a) The actor account is being managed by a marketplace and it needs to sign requests on the actor's behalf
b) The marketplace is providing metadata about the actor (i.e. you need to resolve an endpoint to query metadata about an actor

13/README.md
```diff
@@ -0,0 +1,628 @@
...
+  "jsonrpc": "2.0",
+  "result": "0x8f0227d45853a50eefd48dd4fec25d5b3fd2295e"
+}
+```
+
+The **address** attribute obtained in the response will be **actorId** of the new Actor.
+
+#### Interaction with the Keeper
+
+The KEEPER::Decentralized VM will store the essential user information to allow the implementation of the Actors TCR.
```
mikera: The primary purpose of storing actor information on-chain is to establish actor identity, not to allow a TCR surely?

13/README.md
```diff
@@ -0,0 +1,628 @@
...
+
+Key requirements:
+
+* Only the essential information about Actors will be stored on-chain to allow the implementation of a TCR.
+* AGENT MUST NOT store any information about the accounts (public keys, private keys, passwords or recovery phrases) in any other place than the system KMS.
+* After to be registered, the new Actors MUST have a valid Account Id and a Wallet.
+* If a complementary Ocean DB is given as parameter, the information or Metadata about the Actors will be stored in Ocean DB.
+* The storage of Actors Metadata is totally optional.
+* The location of the KMS SHOULD be provided as parameter in the Ocean AGENT initialization script.
+* The AGENT will use Parity Wallet as technology component to manage the user Wallets.
```
mikera: Seems like an implementation detail rather than a requirement? Perhaps should be moved to the wallet OEP?

13/README.md
```diff
@@ -0,0 +1,628 @@
...
+
+It's out of the scope of this OEP to discuss the Wallet management, this will defined in the scope of the [Wallet OEP](https://github.com/oceanprotocol/OEPs/issues/32).
+
+### Proposed Solution <a name="proposed-solution"></a>
+
+The proposed solution is composed by the interaction of different elements:
+
+* A high level RESTful API exposing the methods required to manage the Actors Registry (AGENT)
+* A Keeper node registering the Actors metadata on-chain (KEEPER)
+
```
mikera: Presumably, if Actor data is going to be stored off-chain then there also needs to be interaction with whatever service providers / marketplaces provide that data?

13/README.md
```diff
@@ -0,0 +1,679 @@
...
+         * [Updating Actor metadata](#updating-actor-metadata)
+         * [Retire an Actor](#retire-an-actor)
+      * [TODO: Events](#events)
+      * [Copyright Waiver](#copyright-waiver)
+      
+<!--te-->
+
+# Ocean Actors Registry <a name="ocean-actors-registry"></a>
+
+The Ocean Actors Registry (**REG**) is a specification for Ocean Protocol to register to all the actors interacting in the Ocean Network.
```
diminator: I don't think we should have a registry to all actors as this could lead to censorship of actors

It makes sense in a specific context, for example giving permissions to actors to engage with other actors (like marketplaces) or specific assets (as permissions set by the asset owners)

13/README.md
```diff
@@ -0,0 +1,679 @@
...
+
+This OEP does not focus on actor attributes, whitelisting or curation. It's purely the mechanics of registering and basic handling of actors.
```
diminator: It should be enough to just have a wallet derived from your KMS to have an account and start transacting and interacting with smart contracts. If specific contracts require permissions, then those permissions should be assigned to the account (which is in essence a public key)

13/README.md
```diff
@@ -0,0 +1,679 @@
...
+The **Actors** information should be managed using an API. This API should exposes the following capabilities:
+
+* Registering an Actor
+* Get an Actor
+* Update an Actor
+* Retire an Actor
+
+Key requirements:
+
+* Only the essential information about Actors will be stored on-chain to allow the implementation of a TCR.
```
diminator: TCRs can be simply account (public key) based - this doesnt require any actor registration

13/README.md
```diff
@@ -0,0 +1,679 @@
...
+* AGENT MUST NOT store any information about the accounts (public keys, private keys, passwords or recovery phrases) in any other place than the system KMS.
+* After to be registered, the new Actors MUST have a valid Account Id and a Wallet.
+* If a complementary Ocean DB is given as parameter, the information or Metadata about the Actors will be stored in Ocean DB.
```
diminator: the actor could publish a DID, which then could be resolved to a self-sovereign DDO (OceanDB, events log, uport, sovrin or another self-sovereign choice)

13/README.md
```diff
@@ -0,0 +1,679 @@
...
+* Registering an Actor
+* Get an Actor
+* Update an Actor
+* Retire an Actor
+
+Key requirements:
+
+* Only the essential information about Actors will be stored on-chain to allow the implementation of a TCR.
+* AGENT MUST NOT store any information about the accounts (public keys, private keys, passwords or recovery phrases) in any other place than the system KMS.
+* After to be registered, the new Actors MUST have a valid Account Id and a Wallet.
```
diminator: Each actor can create a wallet and account off-chain, simply using their wallet software and KMS

diminator: from http://ethdocs.org/en/latest/account-management.html?highlight=address#keyfiles

> Every account is defined by a pair of keys, a private key and public key. Accounts are indexed by their address which is derived from the public key by taking the last 20 bytes.

13/README.md
aaitor: Thinking in a scenario where different marketplaces could have independent TCR's (a data commons marketplace could use a different TCR than a medical data marketplace), it's necessary to associate the Actor to the Marketplace, and initialize the Smart Contract with those basic values (actorId - mktId - status).

13/README.md
aaitor: Details added about DID & DDO

13/README.md
aaitor: Yep, information about how to create wallets offline will be detailed in the Wallet OEP. Reference about this already added.

13/README.md
aaitor: Added clarification about DID

13/README.md
aaitor: The system will store the DID (on-chain) and DDO (blockchain events) allowing to publish data enabling the services discovery (ie. Providers endpoints).

13/README.md
aaitor: Ocean DB only stores Actor Metadata. Protocol doesn't need metadata to work. DID + DDO will provide the mechanisms to locate the actors needing to be discovered.

aaitor:
Actors registry is not part of the Ocean implementation. Archiving this work.

Issue #10, opened 2018-04-16
Ocean Asset Registry (OEP-14)


aaitor:

H34D:
https://github.com/oceanprotocol/OEPs/pull/34

PR #34, opened 2018-05-07
Feature/OEP-14 Assets Registry

aaitor:
PR implementing #10


gongf05 (commented):
The overall OEP looks very reasonable. One issue is the data structs and functions of the smart contracts may need some updates in the future implementation. It's really hard to finalize the design at this stage.  But I believe it is a good version to start with :)

aaitor:
I agree @gongf05 , the definition upfront of the Smart Contract interfaces, parameters, types, structs, etc. is quite difficult before to start. But I hope the current definition helps to show the overall idea, and we will adapt and refine during the development phase.

By the way, if the PR looks okay to you, I suggest to approve it :)
Having some approvals would allow to move on and start working in this development or alternative OEP's.

gongf05 (approved):
Great OEP to start with development.

mikera (changes requested):
Some issues to resolve:

A) The TCR section in here seems like it is a little bit bolted on and incomplete. I suggest moving this to a separate OEP, independent of asset registration?
B) Implementation details seem to be creeping into API specification. Publishers registering assets shouldn't need to know about implementation details like the proposed Ocean DB or the particular solidity contracts used. I suggest keeping the OEP focused on the interface / API contracts
C) Need to be clear precisely which metadata is stored on-chain vs. what is stored elsewhere (related to the issue with 3/4/5). We can afford to detail the specifics of off-chain metadata later or in separate OEPs but the on-chain stuff must be solid with behaviour completely specified.

14/README.md
```diff
@@ -0,0 +1,981 @@
...
+#### Interaction with the Decentralized VM
+
+The **KEEPER::Decentralized VM** will persist the following information:
+
+| Attribute | Type | Description |
+|:----------|:-----|:------------|
+|assetId    |byte32|Asset Id|
+|actorId    |address|Owner of the Asset|
+|contentState      |uint|TCR State of the Asset|
+
```
mikera: This seems insufficient for a consumer to resolve information required to locate asset metadata or the marketplace from which it can be obtained? I think you need some form of marketplace / service provider ID?

14/README.md
```diff
@@ -0,0 +1,981 @@
...
+Using any of the existing web3 implementation library (web3.js, web3.py, web3.j, etc), it's possible to interact with the VM Smart Contracts.
+
+For example using the java Smart Contract stubs generated by web3.j (Java), it's possible to implement a DTO wrapping the Smart Contract interactions. 
+
```
mikera: This seems like implementation detail? Does it need to be in the OEP? I'd expect all users to interact with the Ocean Protocol API as exposed by the Agent, so if we are to provide usage examples in the OEP it should be API calls at that level.

14/README.md
```diff
@@ -0,0 +1,981 @@
...
+|**Metadata**:|     ||
+|creationDatetime   |datetime   |Allocated by the system when was created in the AGENT (universal datetime), time before consensus |
+|updateDatetime     |datetime   |Allocated by the system when was updated the metadata in the AGENT (universal datetime), time before consensus |
+|name       |string|Asset name (optional)|
+|mimeType       |string|The mime-type of the file (optional)|
+|attributes |array |Array of key, value attributes (optional)|
+|parameters |array | If it's a service or operation, specifies the K,V parameters (optional)|
+|links |array |List of links to other assets (samples, previous versions, etc.) (optional)|
+|providers       |object     |Providers giving access to the data. This is managed by the [**ASE.005**](#provider-asset) operation.|
+
```
mikera: I think this OEP should specify the Assets registration API, not implementation details of what happens with the proposed Ocean DB. In separate reviews, I have questioned the design of the Ocean DB approach so I suggest removing these references here pending resolution of broader architectural issues.

14/README.md
```diff
@@ -0,0 +1,981 @@
...
+The Access Control checks if the user requesting to Update the Asset, has enough privileges to do it. 
+To do that, the ```canUpdate(assetId)``` method is called. This method should return a boolean value indicating if the Asset can be modified.  
+
+```solidity
+    function canUpdate(bytes32 _assetId) public returns (bool canUpdate) { }
+    
+    function update(bytes32 _assetId, address _marketplaceId) public returns (bool success) { }
+   
+```
+
```
mikera: I don't think ACL checks should always be on-chain? Better if these can be implemented off-chain, e.g. with calls to the appropriate service provider or marketplace that returns 403 if the ACL check fails.

gongf05:
Updated the OEP-14 smart contracts for Assets Registry to make it consistent with our alpha version implementation.

14/README.md
```diff
@@ -0,0 +1,913 @@
...
+      }
+    ]
+}
+```
+
+
+---
+
+<a name="provider-asset-update"></a><a name="ASE.006"></a>
+### Updating Asset Provider
```
gongf05: in alpha version, the owner is the provider of data asset. There is no need to update asset provider.

aaitor:
This OEP is out of date, closing.

PR #36, opened 2018-05-19
Feature/OEP-06 invoke services

mikera:
Include the WIP version of OEP 6.

This is still a RAW OEP, will continue to extend however good to merge into master branch at this point as it contains key requirements that may need to be referenced by other OEPs.

Agreed with Don that OEP 6 should specify the **interface** for invoking compute services. Different implementations of compute services (Spark, Hadoop, SALPCHAIN etc.) should have separate OEPs and reference this one as needed.


mikera:
Hi @shark8me can you update and merge as WIP?

6/README.md
```diff
@@ -0,0 +1,106 @@
+```
+shortname: 6/CSAPI
```
shark8me: Invoke instead of CSAPI

PR #38, opened 2018-05-22
Transactions OEP 9

aaitor:


9/README.md
```diff
@@ -0,0 +1,803 @@
...
+            "value": "Looking Ahead"
+        }]
+	}
+}
+```
+
+#### Orchestration Layer
+
+The AGENT node will be in charge of manage the Contracts creation. 
+The Ocean DB integration is optional, so the Metadata will be stored there only if Ocean DB interface implementation is provided.
```
mikera: Transactions should not depend on implementation details such as Ocean DB

9/README.md
```diff
@@ -0,0 +1,803 @@
...
+
+<a name="specification"></a>
+## Specification 
+
+Main requirements are:
+
+* TX MUST provide the Smart Contracts logic allowing to non-trusted parties to negotiate the ASSETS sharing
+* Main TX business logic will be implemented in the Smart Contract layer running in the **KEEPER::DEC-VM**
+* The AGENT will interface with the KEEPER::DEC-VM to compose the valid transactions needed by the Smart Contracts layer
+* The AGENT will integrate the KEEPER::DEC-VM to implement the Access Control capabilities
```
mikera: Why should the Keeper need to get involved with ACL? I think it only needs to check once a contract is fully signed? Everything else can be done off-chain.

aaitor:
OEP out of date, closing

PR #54, opened 2018-08-08
Feature/oep 3 on chain auth

aabdulwahed:
Updating the smart contracts to be consistent with the current implementation in keeper.


gongf05:
the smart contract is consistent with our alpha release version.

PR #57, opened 2018-08-09
Feature/oep 14 asset registry refactoring

aabdulwahed:


14/README.md
```diff
@@ -310,7 +334,7 @@ The output of this request MUST add the following attributes generated by the sy
 These are the possible values of the **contentState** attribute:
 
 * PENDING - Asset just created
```
gongf05: In existing alpha version, the data asset has only one "bool status", which denotes data asset is active or not. By default, the status is active when data asset is newly registered. It can be set to be "false" so no one can access it.

14/README.md
```diff
@@ -503,31 +473,20 @@ Internal state attributes and the rest of the attributes can't be modified using
...
-
-* **marketplace**   
-
-In addition to this, integrates the Decentralized VM to implement the access control mechanism.
-The Access Control checks if the user requesting to Update the Asset, has enough privileges to do it. 
-To do that, the ```canUpdate(assetId)``` method is called. This method should return a boolean value indicating if the Asset can be modified.  
-
-If the Asset can be updated, and Ocean DB is enabled, the Orchestration layer will persist the complete Asset metadata in Ocean DB. 
-Also the attribute **updateDatetime** will be updated with the KEEPER universal datetime. 
+#### Updating Asset flow
```
gongf05: there is no functions in Keeper contract to update asset information. But it should be able to be updated in DB.

14/README.md
```diff
@@ -503,31 +473,20 @@ Internal state attributes and the rest of the attributes can't be modified using
...
+The AGENT node will be in charge of manage the Assets update by sending <code>Asset update</code> signal to 
+ocean market contract in order to commit the new <code>metadata hash</code> and keep track the metadata version
+ in smart contract, then make a call to the ocean DB in order to commit the final update. 
 
----
 
 <a name="retire-asset"></a><a name="ASE.004"></a>
 ### Retiring an Asset 
 
 ![Retire an Asset](images/ASE.004.png "ASE.004")
```
gongf05: the only way to retire a data asset is to set "bool status" to be false.

gongf05:
this branch is outdated. Please rebase to get updates from the master branch.

gongf05:
there is some difference between OEP-14 and alpha release. I suggested revising the OEP-14 according to our keeper in alpha release. Thanks.

aaitor:
@gongf05 , thanks for the review. Yep, there are a few things are not totally implemented in Plankton. I like to think this OEP define the complete implementation we need to have about Assets, and Plankton implements the 80%. The plan is continue evolving the implementation in Trilobite release.

14/README.md
aabdulwahed: regarding the status, I prefer to have multiple states as we did in <code>market and authorization</code> contracts but this will be in the next release.

aabdulwahed:
@gongf05 I understand what you mean, but don't worry all these changes will be handled in the next release, the thing I want to do here that we need to work on the same branch and close this refactoring branch, so any minor changes will happen on one branch <code>feature/OEP-14-asset-registry</code>.

aabdulwahed:
@aaitor @gongf05 can I delete this branch? It is no longer needed anymore.

gongf05:
@aabdulwahed it can be deleted.

PR #74, opened 2018-09-12
OEP-08: References to ontology terms compliant to standards

shark8me:
Refactored a few ontology terms to be in line with [DCAT](https://www.w3.org/TR/vocab-dcat/) and Schema.org standards


aaitor:
By the time being the Metadata is using schema.org attributes, and was implemented in that way. Those can be directly translated to DCAT. I guess for changing the current specification and implementation from schema.org to DCAT would be necessary to know the benefits of doing that.

shark8me:
DCAT and Schema.org are largely the same, save for these:

|Schema.org|DCAT|
|-|-|
| name|title|
|inLanguage|language|
|fileFormat|mediaType|

Another issue is that DCAT is covers more of the data catalog ontology compared to Schema.org. So we would have to use Schema defined ontology where defined and then fall back to DCAT when not available. Which isn't consistent. 
In my opinion it should support both fields (e.g. name/title), so that datasets exported from tools such as CKAN (which use name and not title), can be easily exported to Ocean without needing a processing stage to rename missing terms.

mikera:
I don't have a strong view on schema.org vs. DCAT. Both seem to be relatively widely used. DCAT seems a slightly better fit for the domain (dataset metadata) but arguably our assets are a bit broader than that (algorithms etc.)

I do think it is important to however  to pick one approach rather than attempting to create a hybrid: it is better to do this and if necessary map to the other rather than force people to handle both possibilities.

Remember that we will also need some additional ocean-specific metadata fields.

shark8me:
@diminator metadata standards need to balance two concerns:
* allow marketplaces to define their own metadata
* propose a basic "recommended" metadata standard that enables a tooling ecosystem to develop

While marketplaces can choose to completely ignore the recommended metadata, they lose the benefit of tooling. 
Your statement about "allowing marketplaces to have flexibility about choosing metadata standards" also applies to the approved version of OEP-8. In your  opinion is the current set of [base attributes](https://github.com/oceanprotocol/OEPs/tree/master/8#base-attributes), a recommendation, or is it mandatory?

diminator:
Probably marketplaces will set their own recommendations on metadata compatibility.
There are some core attributes required to resolve metadata network wide (like required base attributes), but from a protocol level, what is required are the endpoints to connect to the service agreement:
metadata endpoints, access control endpoints and mechanics and other events to successfully resolve the service agreement.

H34D:
@shark8me @diminator can we close this PR?

shark8me:
Yes, I'll close this one.

PR #75, opened 2018-09-18
Fix/OEP-07 DID

aaitor:


7/README.md
```diff
@@ -52,12 +58,13 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 The main motivations of this OEP are:
 
-* Design a solution to extend the current architecture to use **Uniform Resource Names (URN)**, **Decentralized Identifiers (DID)** and **DID description objects (DDO)**
+* Design a solution to extend the current Ocean architecture to use  **Decentralized Identifiers (DID)** and **DID description objects (DDO)**
 * Understand how to register information on-chain with off-chain integrity associated
 * Understand how to resolve DID's into DDO's
+* Understand how to resolve Assets Metadata using Assets IDs
 * Design a solution facilitating to have the on-chain and off-chain information aligned
-* Establishing the mechanism to know if the DDO associted to a DID was modified
```
aabdulwahed: This is minor change ( we need to follow the W3C specification), they call DDO (DID Document). DDO was the old naming convention.

7/README.md
```diff
@@ -74,16 +81,16 @@ Requirements are:
 * KEEPER doesn't store any ASSET metadata
 * OCEAN doesn't store ASSET files contents
 * SUBJECTS (Ocean actors) could be identified using a **Decentralized ID (DID)** included on-chain and off-chain
-* ACTORS don't need to be registered. Only ACTORS requiring to be discovered by others require a simple registering mechanism
+* ACTORS don't need to be registered. Only ACTORS requiring to be discovered by others require a simple registration mechanism
 * A Decentralized Distributed Object (**DDO**) represents the metadata of an Actor
-* A **DID** can be resolved to get access to a **DDO**
```
aabdulwahed: The current resolving mechanism follows is that each DID will be associated with http link (stored in smart DID registry contract) which points to DID Document.

7/README.md
```diff
@@ -122,71 +129,44 @@ did:ocn:21tDAKCERh95uGgKbJNHYp
...
-Here a draft **DIdRegistry** implementation:
+Here a draft **DidRegistry** implementation:
 
 ```solidity
 
 // This piece of code is for reference only!
 // Doesn't include any validation, types could be reviewed, enums, etc
 
-contract DIdRegistry {
+contract DidRegistry {
```
aabdulwahed: I think "DIDRegistry" is more better :)

7/README.md
aabdulwahed: Sorry, I found it in the description below 👍

7/README.md
```diff
@@ -52,48 +59,55 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
+* ASSET Metadata (off-chain) is associated to the ASSET information stored on-chain using a common **ASSET ID**
+* A **ASSET ID** can be resolved to get access to extended **Metadata**
+* ASSET Metadata could be provided by one or more than one providers
+* ASSETs Metadata can be updated without updating the on-chain information
 * ASSET information stored in the keeper will include a **checksum** attribute
-* The ASSET on-chain checksum attribute, includes a one-way HASH calculated using the DDO content
-* After the DDO resolving, the DDO HASH can be calculated off-chain to validate if the on-chain and off-chain information is aligned
-* A HASH not matching with the checksum on-chain means the DDO was modified without the on-chain update
+* The ASSET on-chain checksum attribute, includes a one-way HASH calculated using the DID Document content
+* After the Metadata resolving, the Metadata HASH can be calculated off-chain to validate if the on-chain and off-chain information is aligned
```
ssallam: `After resolving the Metadata`, ...

7/README.md
```diff
@@ -52,48 +59,55 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
+* A HASH not matching with the checksum on-chain means the DID Document was modified without the on-chain update
 * The function to calculate the HASH MUST BE standard
```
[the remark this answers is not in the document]
aabdulwahed: I think this will be detailed in the asset registry not here

7/README.md
```diff
@@ -52,48 +59,55 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
 ## Proposed Solution
 
 ### Decentralized ID's (DID)
 
-A DID is a unique identifier that can be resolved or de-referenced to a standard resource describing the entity (a DID Document or DDO).
-If we apply this to Ocean, the DID would be the unique identifier of an object represented in Ocean (ie. the Asset ID of an ASSET or the Actor ID of a USER).
-The DDO would be the METADATA information associated to this object that is stored off-chain on Ocean.
+A DID is a unique identifier that can be resolved or de-referenced to a standard resource describing the entity (a DID Document).
+If we apply this to Ocean, the DID would be the unique identifier of an actor interacting in Ocean (the Actor ID of a USER).
+The DID Document would be the METADATA information associated to this Actor DID that is stored off-chain on Ocean.
```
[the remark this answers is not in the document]
aabdulwahed: I agree, here is an example 
```json
{ 
    "@context":{
        "identity":"https://w3id.org/identity/v1#",
        "sec":"https://w3id.org/security/v1#",
        "cred":"https://w3id.org/credentials/v1#",
        "xsd":"http://www.w3.org/2001/XMLSchema#",
        "schema":"http://schema.org#"
     },
    "@type":"identity",
    "id":"did:ether:0x123456789abcdefghi",
     "claim":{
       "zone-1":{
          "@type": ["verifiableCredential", "genericVCProof"],
          "id":"<ServiceEndPointID>",
          "serviceEndpoint":"http://zone1.provider.com/assets",
          "port": 5555,
       }
     }
    ...,
}
```
@aaitor should help in this part

7/README.md
aabdulwahed: done!

7/README.md
```diff
@@ -115,150 +129,212 @@ did:ocn:21tDAKCERh95uGgKbJNHYp
...
-1. A CONSUMER (it could be a frontend application or a backend software), having a DID and using a client library (Python or Javascript) get the **provider** attribute associated to the DID directly from the KEEPER
-1. The CONSUMER, using the provider public url, query directly to the provider passing the DID to obtain the DDO
+1. A PUBLISHER, using the KEEPER, register the new ASSET providing the ASSET URN and the attribute to resolve the provider (DID)
+1. The KEEPER register the ASSET using the OceanMarket Smart Contract. As part of the registerAsset function, a new event is emitted with the Provider DID as Attribute
+1. The PUBLISHER publish the Metadata in the metadata-store/OCEANDB provided by PROVIDER
+1. A CONSUMER (it could be a frontend application or a backend software), having an ASSET URN and using a client library (SQUID libraries) get the **provider** DID attribute associated to the URN directly from the KEEPER
+1. The CONSUMER library (SQUID) using the Provider DID get the provider public url attribute
+1. The CONSUMER, using the provider public url, query directly to the provider passing the URN to obtain the Asset Metadata
```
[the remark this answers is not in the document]
aabdulwahed: done!

7/README.md
```diff
@@ -115,150 +129,212 @@ did:ocn:21tDAKCERh95uGgKbJNHYp
...
 
 
 ## Metadata Integrity
 
 The Metadata Integrity policy is a sub-specification for the Ocean Protocol allowing to validate the integrity of the Metadata associated to an on-chain object (initially an ASSET).
 
-An ASSET in the system is composed by on-chain information maintained by the KEEPER and off-chain Metadata information (DDO) stored in OCEANDB.
-Technically a user could update the DDO accessing directly to the database, modifying attributes (ie. License information, description, etc.) relevant to a previous consumption agreement with an user.
-The motivation of this is to facilitate a mechanism allowing to the CONSUMER of an object, to validate if the DDO was modified after a previous agreement.
+An ASSET in the system is composed by on-chain information maintained by the KEEPER and off-chain Metadata information (DID Document) stored in OCEANDB.
```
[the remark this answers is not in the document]
aabdulwahed: done!

7/README.md
```diff
@@ -52,48 +59,55 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
 
-* The DID resolving capabilities MUST be exposed in the client libraries, enabling to resolve a DDO directly in a totally transparent way
+* The DID resolving capabilities MUST be exposed in the client libraries, enabling to resolve a DID Document directly in a totally transparent way
 * ASSETS are DATA objects describing RESOURCES under control of a PUBLISHER
+* PUBLISHER or OWNERS of ASSETS could delegate some usage permissions regarding those ASSETS
 * KEEPER stores on-chain only the essential information about ASSETS
 * PROVIDERS store the ASSET metadata off-chain
 * KEEPER doesn't store any ASSET metadata
 * OCEAN doesn't store ASSET files contents
+* SUBJECTS (Ocean actors) could be identified using a **Decentralized ID (DID)** included on-chain and off-chain
```
diminator: I think we can define Ocean assets as subjects as well. This would be in the context of self-steering assets:
- the asset signalled itself into the tribe
- the controller of the asset made the asset update it's DDO

diminator: This would allow assets to be treated as a first class citizen in Ocean, much like smart contracts in ETH have the same abilities as wallet owners: they have a balance and can create transactions

diminator: image

diminator: Also this implies that metadata can be wrapped as a DDO attached to the DID of the asset

7/README.md
```diff
@@ -52,48 +59,55 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
+* ASSET Metadata could be provided by one or more than one providers
+* ASSETs Metadata can be updated without updating the on-chain information
 * ASSET information stored in the keeper will include a **checksum** attribute
-* The ASSET on-chain checksum attribute, includes a one-way HASH calculated using the DDO content
-* After the DDO resolving, the DDO HASH can be calculated off-chain to validate if the on-chain and off-chain information is aligned
-* A HASH not matching with the checksum on-chain means the DDO was modified without the on-chain update
+* The ASSET on-chain checksum attribute, includes a one-way HASH calculated using the DID Document content
+* After resolving the Metadata, the Metadata HASH can be calculated off-chain to validate if the on-chain and off-chain information is aligned
+* A HASH not matching with the checksum on-chain means the DID Document was modified without the on-chain update
 * The function to calculate the HASH MUST BE standard
```
diminator: Are we also looking into the storage layer of the DID? similar to identity hubs

7/README.md
```diff
@@ -52,48 +59,55 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
...
-* ASSETS can be resolved using a Decentralized ID (DID) included on-chain and off-chain
-* A Decentralized Distributed Object (**DDO**) represents the ASSET metadata
-* Any kind of object registered in Ocean SHOULD have a DID allowing to uniquely identify that object in the system
-* ASSET DDO (metadata off-chain) is associated to the ASSET information stored on-chain using a common **DID**
-* A **DID** can be resolved to get access to a **DDO**
-* ASSET DDO's can be updated without updating the on-chain information
+* An ASSET is identified using an ASSET ID
+* Any kind of object registered in Ocean MUST have a **Uniform Resource Names (URN)** allowing to uniquely identify that object in the system
+* ASSET Metadata (off-chain) is associated to the ASSET information stored on-chain using a common **ASSET ID**
+* A **ASSET ID** can be resolved to get access to extended **Metadata**
```
diminator: What relevant Metadata would an actor carry?

7/README.md
```diff
@@ -115,150 +129,209 @@ did:ocn:21tDAKCERh95uGgKbJNHYp
...
+### Actors Identity
+
+#### Registry
+
+To register an Actor, requiring to be located, a simple register contract named **DidRegistry** will be provided.
 The key of the Identity entity in Ocean is the **DID**. Associated to this DID we have a Mapping of key-value attributes,
 allowing to associate publicly information to DID's. This could be used to add public information allowing for example
-to discover/resolve a DID.
-There are two main options to implement this:
+to discover/resolve a DID. There are two main options to implement this:
```
diminator: alternatively, the DID registry can just be a tribe's identity hub (ie tribe's oceanDB) that supports query

7/README.md
```diff
@@ -115,150 +129,209 @@ did:ocn:21tDAKCERh95uGgKbJNHYp
...
 // Doesn't include any validation, types could be reviewed, enums, etc
 
-contract IdRegistry {
+contract DIDRegistry {
 
     struct Identity {
         address owner; // owner of the Identity
         string did;
-        unint256 type; // type of Identity object (Asset, Actor, Workflow, ..)
+        mapping (bytes32 => string) attributes; // key => value attributes
```
diminator: what attributes are these?

7/README.md
```diff
@@ -115,150 +129,209 @@ did:ocn:21tDAKCERh95uGgKbJNHYp
...
 }
 
 ```
 
-To register the provider publicly resolving the DDO associated to a DID, we will register an attribute **"provider"** with the public hostname of that provider:
+To register the provider publicly resolving the DID Document associated to a DID, we will register an attribute **"url"** with the public address of that provider:
+
 ```
-registerAttribute("did:ocn:21tDAKCERh95uGgKbJNHYp", "provider", "https://myprovider.example.com")
+registerAttribute("did:ocn:21tDAKCERh95uGgKbJNHYp", "url", "https://myprovider.example.com/")
```
diminator: I'm not sure this is the right approach to query DIDs. In my opinion this is the responsibility of the metadata store. Then there is less public escapes and more privacy control on the metadata. What would make sense is to look at the registry as an integrity check for the DID -> DDO map

aabdulwahed:
@diminator I can see and correct me if I'm wrong, that some assets such as autonomous vehicles could be self-owned due to the intelligence part in the system and at this moment they will interact with the system as  subjects (users/actors). So the intelligence here represents the actor and car body represents the asset. But this isn't valid in case of data, and compute where there is no self-awareness and they should be only assets with owner. Can you explain more how this works?

According to the [DID](https://w3c-ccg.github.io/did-spec/) definition in the w3c specification: <code>"DIDs are fully under the control of the DID subject"</code>. Control here means self-awareness and taking actions

7/README.md
aabdulwahed: From my point of view and as I understand after the conversation with @aaitor, this OEP was meant to be for providers in ocean not all actors. 

- Regarding the provider actor should hold something like public data for the provider such as public keys for the provider, service endpoints, ports, urls, zones (like aws, azure zones) and self signed by the actor's public key. You can think about it as public browser certificate as shown below. A publicly verifiable proof for the provider's metadata and the actor (provider) should have the responsibility to maintain it and keep it available. 

image

7/README.md
aabdulwahed: Yeah, and also they can have their own driver integrated with [universal resolver](https://github.com/decentralized-identity/universal-resolver).

7/README.md
aabdulwahed: @aaitor can you help ?

7/README.md
aabdulwahed: I think about it from different perspective, that users and tribes have identities that are resolved to DID Documents but assets are represented as claims in terms of metadata.

JohnBDB:
@diminator review or close

mikera:
Some of this is superseded by https://github.com/oceanprotocol/OEPs/pull/80 so I would recommend either closing or merging.

diminator:
> Some of this is superseded by #80 so I would recommend either closing or merging.

closing this in favor of #77 
after #77 is merged we can look at #80 and #79

PR #77, opened 2018-09-27
Clarification about metadata as part of the DDO. Small changes in the~Smart Contract

aaitor:
This modification amends the existing OEP-7 spec according to the previous conversation yesterday evening.


7/README.md
```diff
@@ -75,9 +75,9 @@ Requirements are:
 * An ASSET is modeled in OCEAN as on-chain information stored in the KEEPER and metadata stored in OCEANDB
 * ASSETS on-chain information only can be modified by OWNERS or DELEGATED USERS
 * ASSETS can be resolved using a Decentralized ID (DID) included on-chain and off-chain
-* A Decentralized Distributed Object (**DDO**) represents the ASSET metadata
+* A Decentralized Distributed Object (**DDO**) should includes the ASSET metadata
```
diminator: include~s~

mikera:
It is still not clear to me that assets should have DIDs / DDOs. 

The DID model is for self-sovereign identity management - as such it seems that it should apply to actors in the system (service providers etc.) rather than resources (assets etc.). From the DID documentation at https://github.com/WebOfTrustInfo/rebooting-the-web-of-trust-fall2016/blob/master/draft-documents/did-implementer-draft-10.md: 

> "Identity owner. The natural person, party, organization, or thing whose identity is represented by a DID record and who directly controls the private keys to control that record. (Note: this specification avoids the term "user" since an identity owner is not always an individual person.)"

On that basis, I think it is appropriate to use DIDs / DDOs for actors but not assets.

diminator:
> I believe DIDs / DDOs should be reserved for service providers only.
> 
> Asset metadata should be regarded as separate.

We're not looking at the DID spec as being an identity solution only per se
What we want is:
- standards and universal resolvers on the package format
- availability of the metadata

When leaving the metadata to an arbitrary service it violates both requirements and more:
- the metadata api's need to be discoverable, using standard protocol and available
- identity hubs provide availability, consistency and optionally decentralization
- integrity checking based on a standard schema like DIDs
- optionality of creating self-steering assets (smart contracts are for example first class citizens in ethereum)

7/README.md
```diff
@@ -160,20 +198,24 @@ contract IdRegistry {
     constructor(bytes32 _did, uint256 _type) public {
     }
 
-    function registerAttribute(string _did, bytes32 _key, string _value) public returns (bool) {
-        // Option 1.
-        identities[_did][_key] = _value;
+    function register(string _resourceDID, string _providerDID, bytes32 _key, string _value) public returns (bool) {
```
ldmberman: Perhaps, we can keep it flexible and use a generic `registerAttribute` to register both `providerDID` and `providerURL`? Assets, providers, and all other entities can use the same registry contract then.

aaitor: looks a good idea to me. What's drafted is only a very quick reference to illustrate the expected behavior. Those optimizations are much more than welcome!

diminator:
> I believe DIDs / DDOs should be reserved for service providers only.
> 
> Asset metadata should be regarded as separate.

Can we resolve this such that we can base https://github.com/oceanprotocol/OEPs/pull/80 on top of this? @mikera @aaitor

ttmc:
@aaitor asked me to add more details about how to compute a DID, based on a short discussion we had this morning, so that's what I did in commit  https://github.com/oceanprotocol/OEPs/pull/77/commits/caa7cf5e7fe87cc03d0fda2573d9afb000577b75 above.

Note: There were some questions that had to be answered when doing that, so I went with standards or common answers.

I asked @aabdulwahed if he thought it would be okay to let the third part of the DID be 32 bytes long and he said he thinks it should be fine. There's more explanation in the **Length of a DID** section.

mikera:
@ttmc @aaitor I would like to elaborate more on why think there are problems with using some kind of DDO hash as the DID. Hopefully this can clarify:

There is a requirement that identifiers must be stable so that they can be used to address other actors in the Ocean Ecosystem (e.g. a marketplace wishing to explore and index assets on another commons data marketplace would want to use the DID to get a DDO from the universal resolver to get service endpoints for relevant APIs)

How would content-based addressing solve this problem? It would seem that if the DDO changed (e.g. with the addition of a new service endpoint) then you would get a new DID, so it presumably couldn't be used as a stable identifier?

This is effectively an identity management problem, where the task is to maintain the mapping of [ID of a subject] -> [A set of attributes describing the subject]. We need to be able to change the attributes without changing the ID. So I don't see how content-based addressing is useful?

Of course it is fine to use the DDO Hash separately for an integrity check.

Happy to have a discussion offline if that would make more sense.

mikera:
I have approved merging this PR for administrative reasons. Key design issues still need to be addressed but can be done through subsequent PRs.

PR #79, opened 2018-10-02
OEP-07: Proposed DDO hash algorithm

billbsing:


mikera:
I believe we shouldn't be using the DDO hash as a DID, in which case this all appears to be unnecessary.

We can just hash the whole DDO and store next to the DID if the purpose is as an integrity check.

7/README.md
```diff
@@ -234,6 +234,79 @@ The list of changes to apply in the proposed solution are:
 * Integrate the call to the IdRegistry contract in the OceanMarketplace contract - KEEPER
 * Implement the resolving function of a DDO given a DID - CLIENT LIBRARIES
 
+## DDO Hash Algorithm
+
+To reliably hash the contents of a DDO excluding the DDO id.
```
mikera: Suggest the DDO Hash Algorithm should simply be a hash of the full DDO.

diminator:
Please consider first #77 then #80 before we proceed here

diminator:
> I believe we shouldn't be using the DDO hash as a DID, in which case this all appears to be unnecessary.

Why not? Please give me a good reason because this has been blocking for weeks now

billbsing:
I'm just concerned because currently the DDO/Id is not included in the hash. This may have security implications where a single DDO can be re-used for any new newly created DID. 

So whenever a DDO is used, it will also be valid for many other DID's, even when the DDO hash => DID relationship is valid.

diminator:
> I'm just concerned because currently the DDO/Id is not included in the hash. This may have security implications where a single DDO can be re-used for any new newly created DID.
> 
> So whenever a DDO is used, it will also be valid for many other DID's, even when the DDO hash => DID relationship is valid.

How do you mean? It's the basis of content-based addressing, where the ID is the hash of the content. Virtually every decentralized solution employs this strategy for transaction ID, document ID, etc...

diminator:
> I'm just concerned because currently the DDO/Id is not included in the hash. This may have security implications where a single DDO can be re-used for any new newly created DID.
> 
> So whenever a DDO is used, it will also be valid for many other DID's, even when the DDO hash => DID relationship is valid.

How do you mean? It's the basis of content-based addressing, where the ID is the hash of the content. Virtually every decentralized solution employs this strategy for transaction ID, document ID, etc...

billbsing:
I'm not questioning the use of hashing the content. 

I'm questioning the point at which the hashing occurs. When the hash is calculated on the DDO before the ID/DID is applied, then in theory as a provider registering a new DID I can use the same DDO every time I need to generate a DID. 

If it's the other way around, the DID is generated and then the DDO is created using that DID,  then the DDO hash will be unique per DID. 

I have no problems in the hashing of a DDO then using the result to create a DID, I just want to highlight the potential issues that may occur.

billbsing:
I'm not questioning the use of hashing the content. 

I'm questioning the point at which the hashing occurs. When the hash is calculated on the DDO before the ID/DID is applied, then in theory as a provider registering a new DID I can use the same DDO every time I need to generate a DID. 

If it's the other way around, the DID is generated and then the DDO is created using that DID,  then the DDO hash will be unique per DID. 

I have no problems in the hashing of a DDO then using the result to create a DID, I just want to highlight the potential issues that may occur.

billbsing:
I agree with the using content hashing to generate ID's.

I'm more worried about the creator of a DID can use the same DDO and also the subsequently the same DDO hash every time they want to create a new DID.

In the end, I'm ok with using the DDO hash as long as we realise that ID field in the DDO record is pretty much irrelevant and cannot be used for any validation.

mikera:
@dimi 

> > I believe we shouldn't be using the DDO hash as a DID, in which case this all appears to be unnecessary.
> 
> Why not? Please give me a good reason because this has been blocking for weeks now

Two major reasons:
a) It is technically impossible because a standards-compliant DDO *must include* the DID. So you need to allocate the DID first before creating the DDO. See https://w3c-ccg.github.io/did-spec/#did-subject
b) A DID is only really useful if it is a fixed, persistent identifier that others can refer to. You don't want it to change whenever someone updates the DDO (e.g. adding a new service endpoint). It's the same reason you wouldn't want your phone number to change every time you move near a different cell tower :-)

Let me know if you would like to discuss more, but my proposed solution is to close this PR and select an alternative method of DID allocation (ethereum address could be an easy one....)

mikera:
@dimi 

> > I believe we shouldn't be using the DDO hash as a DID, in which case this all appears to be unnecessary.
> 
> Why not? Please give me a good reason because this has been blocking for weeks now

Two major reasons:
a) It is technically impossible because a standards-compliant DDO *must include* the DID. So you need to allocate the DID first before creating the DDO. See https://w3c-ccg.github.io/did-spec/#did-subject
b) A DID is only really useful if it is a fixed, persistent identifier that others can refer to. You don't want it to change whenever someone updates the DDO (e.g. adding a new service endpoint). It's the same reason you wouldn't want your phone number to change every time you move near a different cell tower :-)

Let me know if you would like to discuss more, but my proposed solution is to close this PR and select an alternative method of DID allocation (ethereum address could be an easy one....)

mikera:
@dimi 

> > I believe we shouldn't be using the DDO hash as a DID, in which case this all appears to be unnecessary.
> 
> Why not? Please give me a good reason because this has been blocking for weeks now

Two major reasons:
a) It is technically impossible because a standards-compliant DDO *must include* the DID. So you need to allocate the DID first before creating the DDO. See https://w3c-ccg.github.io/did-spec/#did-subject
b) A DID is only really useful if it is a fixed, persistent identifier that others can refer to. You don't want it to change whenever someone updates the DDO (e.g. adding a new service endpoint). It's the same reason you wouldn't want your phone number to change every time you move near a different cell tower :-)

Let me know if you would like to discuss more, but my proposed solution is to close this PR and select an alternative method of DID allocation (ethereum address could be an easy one....)

mikera:
> > I believe we shouldn't be using the DDO hash as a DID, in which case this all appears to be unnecessary.
> 
> Why not? Please give me a good reason because this has been blocking for weeks now

@dimi  two major reasons:
a) It is technically impossible because a standards-compliant DDO *must include* the DID. So you need to allocate the DID first before creating the DDO. See https://w3c-ccg.github.io/did-spec/#did-subject
b) A DID is only really useful if it is a fixed, persistent identifier that others can refer to. You don't want it to change whenever someone updates the DDO (e.g. adding a new service endpoint). It's the same reason you wouldn't want your phone number to change every time you move near a different cell tower :-)

Let me know if you would like to discuss more, but my proposed solution is to close this PR and select an alternative method of DID allocation (ethereum address could be an easy one....)

mikera:
> > I believe we shouldn't be using the DDO hash as a DID, in which case this all appears to be unnecessary.
> 
> Why not? Please give me a good reason because this has been blocking for weeks now

@dimi  two major reasons:
a) It is technically impossible because a standards-compliant DDO *must include* the DID. So you need to allocate the DID first before creating the DDO. See https://w3c-ccg.github.io/did-spec/#did-subject
b) A DID is only really useful if it is a fixed, persistent identifier that others can refer to. You don't want it to change whenever someone updates the DDO (e.g. adding a new service endpoint). It's the same reason you wouldn't want your phone number to change every time you move near a different cell tower :-)

Let me know if you would like to discuss more, but my proposed solution is to close this PR and select an alternative method of DID allocation (ethereum address could be an easy one....)

ldmberman:
> a) It is technically impossible because a standards-compliant DDO must include the DID. So you need to allocate the DID first before creating the DDO.

If DID is computed as a hash of the DDO fields other than DID itself (what is proposed here), it complies with the spec then.

> b) A DID is only really useful if it is a fixed, persistent identifier that others can refer to. You don't want it to change whenever someone updates the DDO.

I think, we need to be specific about how these primitives are gonna be used.

I can imagine clients which do not want to bother watching for DDO updates but would rather consult a registry providing it with the DID. Then, in order to serve updated documents by the old identifiers, such a registry would need to keep track of the identifiers across the updates. Such an approach is possible but is more cumbersome than using a fixed identifier indeed.

diminator:
> a) It is technically impossible because a standards-compliant DDO _must include_ the DID. So you need to allocate the DID first before creating the DDO. See https://w3c-ccg.github.io/did-spec/#did-subject

please read the spec #77 first



> b) A DID is only really useful if it is a fixed, persistent identifier that others can refer to. You don't want it to change whenever someone updates the DDO (e.g. adding a new service endpoint). It's the same reason you wouldn't want your phone number to change every time you move near a different cell tower :-)

different DDO, different DID

H34D:
can we close this PR?

PR #80, opened 2018-10-11
Update OEP-07 to use DIDs for Identities, DID Paths for Assets

mikera:
The current OEP suggests that DID and DDOs should be registered for both Actors and Assets. This is a problem for serveral reasons:

- Assets are resources, not identities therefore are not a good fit for an Identity Management approach (they do not in general own their own public and private keys for example)
- Registering all Assets on-chain would be prohibitively expensive, given that a publisher may need to register thousands or even millions of Assets. 

This PR resolves these problems by:
- Retaining the DID/DDO approach for Actors / service providers
- Addressing Assets as DID paths under the DID of the relevant service provider (this is as recommended by the draft W3C spec: "A DID path SHOULD be used to address resources available via a DID service endpoint")


mikera:
Hi @aaitor , can you review this PR, see if you agree and if so merge? Thanks!

aaitor (changes requested):
This was discussed a couple of weeks ago, also there is another PR about DID's with a different approach and different opinions.
As was said, we understand DID as a valid mechanism to identify not only Actors, also Assets.

mikera:
@Aitor this proposal does allow assets to be referenced via the DID Path approach recommended in the W3C spec, e.g. 

`did:ocn:cd2a3d9f938e13cd947ec05abc7fe734df8dd826/c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470`

Can you point me to the alternative proposal? Is it this PR? https://github.com/oceanprotocol/OEPs/pull/75

7/README.md
```diff
@@ -33,10 +33,29 @@ Table of Contents
 
 # Decentralized Identifiers
 
-This OEP doesn't detail the exact method of registering ASSETS on-chain or publishing metadata in a metadata store.
+This OEP presents an approach to using the W3C DID specification within Ocean for the following purposes:
+- Providing decentralised identifiers (DIDs) for Actors on Ocean
+- Providing way to address Assets as resources managed under the DID of a relevant Actor
```
diminator: Assets can be viewed as a DID with guardianship
https://w3c-ccg.github.io/did-spec/#authorization-and-delegation

Assets are first-class citizens in the Ocean Network (~similar to smart contracts in ETH): 
- they are account holding (through Service Agreements)
- they can have an owner, but not need to (for self-steering assets and commons)

7/README.md
```diff
@@ -33,10 +33,29 @@ Table of Contents
...
+- Providing the ability to resolve a DDO (on-chain) for relevant Actors (especially service providers who need to expose API endpoints for the ecosystem
+- Providing a way to cryptographically verify metadata for assets 
 
-This specification is based on [Ocean Protocol technical whitepaper](https://github.com/oceanprotocol/whitepaper), [3/ARCH](../3/README.md) and [4/AGENT](../4/README.md).
+A DID for an Actor in Ocean takes the follwoing format:
```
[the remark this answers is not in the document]
diminator: ~for and Actor~
I would remove this throughout the whole document

7/README.md
```diff
@@ -33,10 +33,29 @@ Table of Contents
 
 # Decentralized Identifiers
 
-This OEP doesn't detail the exact method of registering ASSETS on-chain or publishing metadata in a metadata store.
+This OEP presents an approach to using the W3C DID specification within Ocean for the following purposes:
+- Providing decentralised identifiers (DIDs) for Actors on Ocean
+- Providing way to address Assets as resources managed under the DID of a relevant Actor
+- Providing the ability to resolve a DDO (on-chain) for relevant Actors (especially service providers who need to expose API endpoints for the ecosystem
+- Providing a way to cryptographically verify metadata for assets
```
diminator: I think metadata could be regarded as a service in the ocean network. Similar as storage, compute, ...

7/README.md
```diff
@@ -33,10 +33,29 @@ Table of Contents
 
 # Decentralized Identifiers
 
-This OEP doesn't detail the exact method of registering ASSETS on-chain or publishing metadata in a metadata store.
+This OEP presents an approach to using the W3C DID specification within Ocean for the following purposes:
+- Providing decentralised identifiers (DIDs) for Actors on Ocean
+- Providing way to address Assets as resources managed under the DID of a relevant Actor
+- Providing the ability to resolve a DDO (on-chain) for relevant Actors (especially service providers who need to expose API endpoints for the ecosystem
```
diminator: Change Actor through Identity throughout the whole document

7/README.md
```diff
@@ -33,10 +33,29 @@ Table of Contents
...
+- Providing a way to cryptographically verify metadata for assets 
 
-This specification is based on [Ocean Protocol technical whitepaper](https://github.com/oceanprotocol/whitepaper), [3/ARCH](../3/README.md) and [4/AGENT](../4/README.md).
+A DID for an Actor in Ocean takes the follwoing format:
 
+`did:ocn:cd2a3d9f938e13cd947ec05abc7fe734df8dd826`
+
+Where the hexadecimal ID is the the ethereum account address of the Actor
```
diminator: Why the account and not a hash of the DDO (excluding the DID)

diminator: It will be important to see how to marshall the information that generates the unique identifier for integrity purpose

As a reference: this is how we compute the ID of a BigchainDB transaction:
https://github.com/bigchaindb/BEPs/tree/master/13

7/README.md
```diff
@@ -33,10 +33,29 @@ Table of Contents
...
+
+An Asset with Metadata provider by an Actor can the be addressed in the following format:
+
+`did:ocn:cd2a3d9f938e13cd947ec05abc7fe734df8dd826/c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470`
+
+Where the two hexadecimal IDs are the ethereum account address of the Actor managing the Asset and ID of the Asset. The ID
```
diminator: We'll first want to figure out what are identities on the ocean network before tying them to ethereum account addresses

7/README.md
```diff
@@ -33,10 +33,29 @@ Table of Contents
...
+`did:ocn:cd2a3d9f938e13cd947ec05abc7fe734df8dd826`
+
+Where the hexadecimal ID is the the ethereum account address of the Actor
+
+An Asset with Metadata provider by an Actor can the be addressed in the following format:
+
+`did:ocn:cd2a3d9f938e13cd947ec05abc7fe734df8dd826/c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470`
+
+Where the two hexadecimal IDs are the ethereum account address of the Actor managing the Asset and ID of the Asset. The ID 
+of the asset in turn is defined as the Hash (Keccak256?) of the Asset Metadata.
```
diminator: Or use multihash: https://github.com/multiformats/multihash

7/README.md
```diff
@@ -33,10 +33,29 @@ Table of Contents
...
+
+This scheme presents the following useful properties:
+- It is possible to address the same Asset provided by multiple Service Providers (e.g. an asset is listed on multple marketplaces for sale).
```
diminator: This would be solved by Service Agreements where there is a 1-to-many relationship between an Asset identity and service agreements

7/README.md
```diff
@@ -33,10 +33,29 @@ Table of Contents
...
+- It is possible to extend this scheme to sub-Assets by extending the DID path
```
diminator: This is not true: the DID probably needs to be limited in length

7/README.md
```diff
@@ -33,10 +33,29 @@ Table of Contents
...
+- There is no requirement for Assets to be registered on-chain, allowing for efficient off-chain registration of Assets and associated Metadata
```
diminator: The main thing that's connecting assets to on-chain are the Service Agreements.
Metadata is regarded as a service

aaitor:
The proposal is OEP-7 and the PR to update this https://github.com/oceanprotocol/OEPs/pull/77

7/README.md
```diff
@@ -52,14 +71,11 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 The main motivations of this OEP are:
 
-* Design a solution to extend the current architecture to use **Decentralized Identifiers (DID)** and **DID description objects (DDO)**
-* Understand how to register information on-chain with off-chain integrity associated
-* Understand how to resolve DID's into DDO's
-* Design a solution facilitating to have the on-chain and off-chain information aligned
-* Establishing the mechanism to know if the DDO associted to a DID was modified
+* Define a standard to identify Actors in Ocean using **Decentralized Identifiers (DID)**
```
diminator: ~Actors~ Identities

7/README.md
```diff
@@ -69,21 +85,14 @@ Requirements are:
 * The DID resolving capabilities MUST be exposed in the client libraries, enabling to resolve a DDO directly in a totally transparent way
 * ASSETS are DATA objects describing RESOURCES under control of a PUBLISHER
 * KEEPER stores on-chain only the essential information about ASSETS
-* PROVIDERS store the ASSET metadata off-chain
+* MATA AGENTS store the ASSET metadata off-chain
```
diminator: META

diminator (changes requested):
This is a good start, but it's clear we need to have convergence on the following:
- abstract Identities that can have an DID (and not only look at actors, but also tribes and assets)
- consider the metadata network as a service network (ie a public or permissioned asset)
- connection between Identities and on-chain would be contracts such as Service Agreements, Verification and Curation Contracts

7/README.md
mikera: I believe DIDs for actors should be stable over time for a provider, i.e. you keep the same DID even if you update the DDO. On this basis, using the Ethereum account makes more sense than the DDO hash (which will change.... and technically should contain the DID anyway according to the spec)

diminator: False: A DDO might change while being versioned, and this would mean the Identity changed. Still the DDO related to that DID will exist. It decouples the account-dependency from the identity

diminator: You can make a DDO with an ethereum account, marshal, compute the DID and store the record.
This would be exactly the same operation. Event better, you would be able to connect multiple keys (ETH, BTC, ...) to a single DID and hence decouple from the dec. VM layer

7/README.md
mikera: Agreed. This does not appear to substantively change the design, and better to refer to Identities to avoid confusion with other Actors.

7/README.md
mikera: Agreed. See the separate OEP (not yet numbered, but potentially OEP 15?) used to define the Meta API. https://github.com/oceanprotocol/OEPs/tree/feature/oepxxx-meta-api. @billbsing or myself will send a separate PR for this.

7/README.md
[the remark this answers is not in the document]
mikera: Agreed

7/README.md
mikera: I suggest this as an approach because:
a) Any Actor wishing to register an Identity with a DID/DDO will need an Ethereum account anyway to pay the gas cost
b) It gives us a mechanism via verification of Ethereum transactions to control updates of DIDs/DDOs
c) It avoids the need to allocate different IDs



A different mechanism for allocating DIDs (e.g. hashing the initial version of the DDO without the DID) would be possible but don't see a motivating reason to add additional complexity here? There are issues with this approach:
- It won't be useful to verify the DDO directly because the DDO may be updated later
- Using the hash of the full DDO is technically impossible if we want to comply strictly with the DID spec because the DDO needs to contain the DID.
- There could potentially be DID collisions, e.g. multiple identities registering the same DDO (e.g. with nothing but identical endpoints)

7/README.md
mikera: Added some proposed wording to define identities in this context. Other suggestions welcome

7/README.md
mikera: It there a motivating use case? It will keep implementations simpler if we can standardised on a hash function (this is mentioned as one of the requirements). I think an Ocean standard is helpful here.

7/README.md
mikera: Provisionally agreed - assuming that the Service Agreement will need to specify the tribe / meta agent in any case so that provenance and metadata of the asset can be obtained? Is there a separate OEP for this?

7/README.md
mikera: Current DID spec does not prescribe a limitation on length. We should probably implement an Ocean standard e.g. 2k. See e.g. https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers

This should provide a standard upper bound without placing too much restriction on nested asset structure (about 30 levels deep if encoded in hex).

Have added a proposed 2k limit in the PR.

7/README.md
mikera: Fixed

7/README.md
diminator: :100:

7/README.md
diminator: I would propose the following for constructing a DDO and related DID:

a. fill in the fields related to the identity. `signature` and `did` fields are empty
b. compute the `signature` field: ie. sign the DDO after (a) and fill in the `signature` field
c. compute the `did` field: marshall and hash the DDO after (b) and fill in the `did` field

This way:
- the DID is an integrity check for the DDO
- the DDO is signed and cannot be replayed by non-holders of the private key
- the DDO and DID are independent of the underlying decentralized VM

an similar example can be found in the BigchainDB docs: https://github.com/bigchaindb/BEPs/tree/master/13#how-to-construct-a-transaction

7/README.md
diminator: Under construction:  https://github.com/oceanprotocol/dev-ocean/blob/sla/version_01/doc/sla.md

The current thinking is that a Service Agreement is composed of conditions. and tribes will bring their own conditions (ie membership, permissions, payment, ...)

image

7/README.md
diminator: If the DID is being used in Smart Contract languages like solidity, maybe it's more useful to look at `bytes32`?

https://solidity.readthedocs.io/en/v0.4.24/types.html#fixed-size-byte-arrays

diminator: Of course we'll have to consider the chances on collisions when truncating, probably this can be overcome to some extent by hashing

7/README.md
```diff
@@ -110,9 +134,14 @@ idchar             = ALPHA / DIGIT / "." / "-"
...
 ```text
-did:ocn:21tDAKCERh95uGgKbJNHYp
+did:ocn:cd2a3d9f938e13cd947ec05abc7fe734df8dd826
 ```
 
+As per section 3.3 of the DID spec (https://w3c-ccg.github.io/did-spec/#paths): "A DID path SHOULD be used to address resources available via a DID service endpoint" therefore we use DID paths to address Assets managed by the relevant Actor.
+
+As per section 9.1 of the DID spec there is no upper limit on DID length, but in Ocean we elect to limit
+DIDs (including DID paths) to 2000 characters to ensure maximum interoperability with clients / URLs.
+
```
diminator: we'd probably can set limits on the DID root and paths separately

diminator (changes requested):
I think we're closing in: main thing to resolve is how to construct the DID

7/README.md
```diff
@@ -33,10 +34,43 @@ Table of Contents
 
 # Decentralized Identifiers
 
-This OEP doesn't detail the exact method of registering ASSETS on-chain or publishing metadata in a metadata store.
+This OEP presents an approach to using the W3C DID specification within Ocean for the following purposes:
+- Providing decentralised identifiers (DIDs) for Identities on Ocean
+- Providing way to address Assets as resources managed under the DID of a relevant Identity
```
diminator: Assets as DIDs with a guardian:
https://w3c-ccg.github.io/did-spec/#authorization-and-delegation

diminator (changes requested):

7/README.md
mikera: Makes sense. Though perhaps you don't actually need to use deeply nested paths anyway: a read from `did:ocn:[provider ID]/[asset ID]/[sub-asset ID]` can probably just be compacted to `did:ocn:[provider ID]/[sub-asset ID]` since the hashing process effectively guarantees the sub-asset will be unique.

7/README.md
mikera: It seems assets are best modelled as "resources managed by an identity" (W3C DID spec 3.3: "A DID path SHOULD be used to address resources available via a DID service endpoint") rather than as entities which have delegated control to another entity (which would be more applicable e.g. in case of legal guardianship, or key recovery as suggested by W3C DID spec 4.5)

Does that make sense? Or do you have an alternative interpretation / design to suggest?

mikera:
@aitor @dimi What is blocking this from being merged?

Do you have additional concrete feedback or suggestions?

7/README.md
```diff
@@ -34,13 +35,45 @@ contributors: Dimitri De Jonghe <dimi@oceanprotocol.com>, Troy McConaghy <troy@o
...
-* the [Ocean Protocol technical whitepaper](https://github.com/oceanprotocol/whitepaper),
-* [3/ARCH](../3/README.md), and
-* [4/AGENT](../4/README.md).
+For the purposes for this OEP, an Identity can be any entity that wishes to register a self-sovereign DID/DDO on
+the Ocean network. Currently this is anticipated to include the following:
+- Any Tribe that wishes to manage metadata for Assets, which is required to offer the Meta API
+- Any off-chain service that needs to provide endpoints for Ocean Agent APIs (including marketplaces, service providers etc.)
+- Any other Actor with an ethereum account who wishes to provide a DDO on Ocean
 
+A DID/DDO is **mandatory** for any Identity wishing to offer Ocean Agent APIs to the ecosystem, because
```
diminator: Especially for assets DIDs in Service agreements

image

7/README.md
```diff
@@ -34,13 +35,45 @@ contributors: Dimitri De Jonghe <dimi@oceanprotocol.com>, Troy McConaghy <troy@o
...
+A DID for an Identity in Ocean takes the following format:
+
+`did:op:cd2a3d9f938e13cd947ec05abc7fe734df8dd826`
+
+Where the hexadecimal ID is a unique ID for the Identity
+
+TODO: consider alternative way to allocate IDs to Identities, could include:
+- Psuedo-random generation
+- Use of Ethereum account
+
```
diminator: - a hash of the DDO

7/README.md
```diff
@@ -303,11 +349,35 @@ The list of changes to apply in the proposed solution are:
 * Implement the resolving function of a DDO given a DID - CLIENT LIBRARIES
 
 
+## Metadata Integrity
```
diminator: This would require a dedicated field in the DDO?

7/README.md
```diff
@@ -292,6 +312,32 @@ Steps:
 1. A CONSUMER (it could be a frontend application or a backend software), having a DID and using a client library (Python or Javascript) get the **service-ddo** attribute associated to the DID directly from the KEEPER
 1. The CONSUMER, using the provider public url, query directly to the provider passing the DID to obtain the DDO
 
+### Ocean standard endpoints
```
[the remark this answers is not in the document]
mikera: Agreed - I will move to a separate OEP

7/README.md
mikera: We still have a requirement for a stable identifier, so using the hash of the DDO in general doesn't work as previously explained.

I agree the hash of the initial DDO (minus ID) could be used to generate an initial value for the stable identifier, but this seems pretty arbitrary and could have collisions. Have included as an an option - but we need to pick one soon!

7/README.md
mikera: I don't believe so. We propose to identify assets as follows:

`did:op:[Provider ID]/[Asset ID]`

Given the above string, you can validate the integrity of the asset metadata using the Asset ID. There isn't a need for anything in the DDO itself.

7/README.md
sawlag: It might be worth ensuring these are a strict sub-set of the more general capabilities for this stream on data sharing agreement (DSA): 
https://spiral.imperial.ac.uk/handle/10044/1/21871  https://www.iit.cnr.it/sites/default/files/analyser.pdf
https://kar.kent.ac.uk/65817/1/An%20Architecture%20for%20Privacy-preserving%20Sharing%20of%20CTI%20with%203rd%20party%20Ana....pdf

The idea would be that the Ocean SAs are a part of wider potential attribute-based access control schemes (implemented by providers and in particular meeting various privacy needs).

Alignment with a strict subset of XACLM  might also be a good ploy: https://en.wikipedia.org/wiki/XACML .

kremalicious:
Seems abandoned, requested changes not provided until now, so closing.

PR #82, opened 2018-10-16
OEP-08: Asset Metadata proposed changes

mikera:
This PR makes a number of proposed improvements to the Asset metadata definition.

- Improve specification for Links to allow linked Assets and metadata on
link types (this is needed to allow Marketplaces to identify sample files etc.)
- Change "tags" attribute to an array (avoid additional string parsing)
- Separate attributes for invokable services vs. datasets (different attributes are relevant)
- Remove curation and pricing from Asset Metadata. These values should not be part of the Asset Metadata as the metadata needs to be immutable.
- Remove references to DDOs as this is an orthogonal concern covered in OEP7


mikera:
@aaitor can you review? I would like to get this merged assuming the general direction is agreed so we can work on further enhancements

mikera:
Closing as relevant changes have been moved to OEP13

PR #83, opened 2018-10-16
Feature/OEP-15 meta api

billbsing:
Mike & I are not sure of the OEP No. ?

First draft of a more generic protocol to add/retrieve metadata in the Ocean ecosystem.


aaitor:
The current Provider component exposes a HTTP API, that it's integrated with Pleuston frontend application. This API is documented and can be reviewed using Swagger.
Is the intention of this to document this Provider existing functionality in an OEP? 
Is this aligned with the existing API?

billbsing:
This is a start of the process to move the `Provider` into separate independent modules, that the API can be supported by one server, or potentially can be supported by individual 'Provider' serverlets. 

So this OEP is to defined as part of the `Provider` server, to replace the '/metadata/<asset_id>' base API's. Where the <asset_id> is the equal to the <metadata_hash>.

The API metadata ```delete``` and ```update``` will have to be removed to allow for a full providence a path of the asset metadata.

The API '/metadata/consume/<asset_id>' should be moved to separate section for an asset based API, within the `Provider`.

aaitor:
@eruizgar91 , can you take a look at it please?
Maybe you can help to @billbsing to unify this API exposed using as input the current implementation.

mikera:
Just to be clear, we need to standardised this if alternative Metadata Providers (e.g. https://github.com/oceanprotocol/surfer which crawls CKAN repositories) wish to provider metadata to Ocean in an interoperable way.

mikera:
@aitor can you point me to where your current implementation is documented with swagger? I will take a look and amend this PR if appropriate.

kremalicious:
You need to have `aquarius` running locally, then the Swagger docs are exposed under http://localhost:5000/api/v1/docs

https://github.com/oceanprotocol/aquarius/blob/develop/docs/for_api_users/README.md

mikera:
If I read correctly from: 

https://github.com/oceanprotocol/aquarius/blob/develop/docs/for_api_users/API.md

The endpoint is an Aquarius-specific API:
`/api/v1/aquarius/assets/metadata/{id}`

This is meant to be a standard API for any metadata provider, not Aquarius specific, so it fulfils a different role and should not have the provider implementation name baked into the URL. I suggest we merge (and if necessary refactor things later).

PR #84, opened 2018-10-23
OEP-07: new contract for resolver with block number as an index to the latest event log

billbsing:
Getting a bit ahead of myself..

As part of the process of getting the Universal DID Resolver working, I was testing out realistic methods in resolving DID's off the block chain, so ....

I want to get the latest DID event log from the block chain without doing the following:

1. Search the whole block chain for an event.

1. Behind the Squid library, having to a run a background process that is always listening for events, and then updating a local cache database for detected events.

So I have updated the spec for the DID resolver to include the following:

1. Saving the Block Number of the event, so this can be found without searching the whole block chain

1. Changing the ```type``` to show the type of resolving data ( e.g. DID, DIDRef, DDO, URL e.t.c. ), instead of the resolving service, as this can be found out once the service is contacted.

### DIFF
--- a/7/README.md
+++ b/7/README.md
@@ -4,7 +4,7 @@ name: Decentralized Identifiers
 type: Standard
 status: Raw
 editor: Aitor Argomaniz <aitor@oceanprotocol.com>
-contributors: 
+contributors:
 ```
 
 <!--ts-->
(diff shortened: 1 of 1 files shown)


aaitor:
IMO this is the summary of introducing this approach.
PROS:
- More efficient filtering on events. Events are indexed in the VM ([Bloom filter](https://ethereum.stackexchange.com/a/3426))so I'm not sure about how much we are improving with this

CONS:
- More storage. One additional uint per entry
- More requests to the Keeper (Smart Contract). Before each filtering, it's necessary to retrieve the blockNumber that it's on-chain

aaitor (changes requested):
Old PR, IMO can be closed

kremalicious:

PR #87, opened 2018-10-27
Allow omission of leading zeroes in DID idstring

ttmc:
Goals of this pull request were to:

- Make it clear that the leading zeroes of the hex string part of the DID can be omitted
- Give some example Ocean Protocol DIDs
- Give a regular expression that Ocean Protocol DIDs must match
- Fix the OEP-7 ToC

I'm actually not 100% sure we want to allow the omission of leading zeroes in the hex string, because that would open the door for flawed DID generators to always generate hex strings that are too short (e.g. always 32 characters long), and to get away with it.

### DIFF
--- a/7/README.md
+++ b/7/README.md
@@ -21,9 +21,7 @@ contributors: Dimitri De Jonghe <dimi@oceanprotocol.com>, Troy McConaghy <troy@o
          * [DID Documents (DDO)](#did-documents-ddo)
          * [Integrity](#integrity)
             * [Length of a DID](#length-of-a-did)
-            * [How to compute a DID for a DDO](#how-to-compute-a-did-for-a-ddo)
-               * [Computing the Proof](#computing-the-proof)
-               * [Computing the SHA-3 Hash](#computing-the-sha-3-hash)
+            * [How to Compute a DID](#how-to-compute-a-did)
          * [Registry](#registry)
          * [Resolver](#resolver)
       * [Changes Required](#changes-required)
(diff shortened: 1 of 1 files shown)


7/README.md
```diff
@@ -164,18 +162,31 @@ It would be nice to store the "did:op:" prefix in those 32 bytes, but that means
...
+The following are examples of valid Ocean Protocol DIDs:
+
+```text
+did:op:9050a77f4acffd30a2e97bfa5abfcb9256f6f9cde60091bf9d573c534052d9fd
+did:op:000000000000000000000bfa5abfcb9256f6f9cde60091bf9d573c534052d9fd
+did:op:bfa5abfcb9256f6f9cde60091bf9d573c534052d9fd
+```
+
+An Ocean Protocol DID string should match the regular expression:
+
```
billbsing: Can we add A-F to allow a bit more flexibility and a possible checksum of id, similar to Ethereum address checksum.

         `^did:op:[0-9A-Fa-f]{1,64}$`

billbsing (changes requested):

billbsing (changes requested):
I have also thinking about the usefulness of a '0' or NULL address.

So  the address 

```did:op:0``` 

is a valid DID address, as it meets the regular expression criteria as a valid address. 

But this address cannot be assigned or owned by any account/actor or used in the DID resolver to resolve to any URL, DDO or DID. It's sole purpose is to resolve to _None/Empty Address_.
As we will be using the  hex part as an Id number for Assets, Provides, e.t.c. A 0 Id will also be _None/Empty_.

If this is the case we will need to do PR's on the Smart Contracts to stop any one owning a 0, address, or id.

mikera:
Don't have a particularly strong view on this, but usually it is helpful to be relatively strict in APIs to avoid edge cases and ambiguity.

Omitting leading zeros has a few potential issues:
- It messes up lexicographical sort order, e.g. "did:op:01000" sorts before "did:op:100". Do we really want this?
- It makes some IDs ambiguous, i.e. you can no longer use toLowerCase(did1)==toLowerCase(did2) as a comparison of did values - you have to parse and compare the values.
- It doesn't actually save anything in terms of storage most of the time: on average you save about 0.07 characters assuming (pseudo) random IDs, which is of negligible value given the extra parsing complexity.

For null/ zero values I would suggest actual  `null` values  (or empty lists) rather than forcing consumers to parse the DID and compare to zero. Otherwise you get all kinds of bugs where a consumer assumes a non-null DID must be valid.

mikera (changes requested):
Request to merge PR 80 first, and base additional changes on top of that

7/README.md
ttmc: I'm okay with allowing A-F.

Regarding the checksum: How sophisticated would you want it to be? Would a single parity bit (0 or 1) be okay? Where would it go? Tacking it on the end of the id string would open the door to all sorts of nasty implementation errors. Maybe put it in a new field in the DDO? That would be okay I guess. (Future versions of Ocean DIDs/DDOs, allowing for DID = hash(DDO without DID) would not need a checksum because the hash/DID itself could be thought of as a kind of checksum of the DDO.)

ttmc:
I'll just wait to see if the devs can come to some consensus and I'll document that.

PR #89, opened 2018-11-01
New OEP for Agent API endpoints

mikera:
This PR creates a new OEP for standardised Ocean service endpoints that may be offered by Ocean Agents.

This is necessary to ensure that clients are able to locate service endpoints for Ocean Agents that they may wish to interact with (e.g. a marketplace).

This has been separated out from OEP 7 (DIDs) because it is a separate concern, however the mechanism for specifying the service endpoints still depends on the OEP 7 DID / DDO approach.

Temporary number of OEP19 has been used, we assume this will be re-allocated


mikera:
@shark8me can you review and merge if OK? Thanks!

diminator:
This seems a bit odd. Assuming we make abstraction of web2 and web3 services by introducing access control (BRIZO) and adapters (OSMOSIS). 

Here's a diagram that would clarify how we model the network. Can you point out how this proposal fits in?

image

PR #90, opened 2018-11-06
OEP-08: Asset Bundles

mikera:
In many cases, assets are naturally composed of multiple data files. Examples:
- Division of machine learning data into test / training / validation data sets
- Individual image files in a resource bundle
- Supporting files e.g. domain-specific metadata and lookup tables

This PR makes it possible for Ocean assets to represent a bundle of other Ocean assets. Used recursively, this creates a merkle tree resembling a directory structure of all assets contents.

Some relevant features of this solution:
- Sub-assets are referenced by the Asset ID (keccak256 hash of metadata)
- Sub-assets may be safely referenced in multiple bundles since they are immutable
- Validation of the integrity of the full tree is possible by validating the relevant hashes at each level


kremalicious:
just some styling and structure for `links`: can we please keep the structure we already have for the `links` in current `master`? No need to prioritize them over all the other base attributes by giving them an extra section, the table explanation is enough and the full example then shows them. Likewise, the example at the end should follow the same structure, and not show a different structure than in the redundant `links` section, just like in current `master`. Right now, this PR shows 2 different `links` structures, and the already established example in `master` has been replaced. Why?

In general, this PR includes way more changes than just adding asset bundles, it seems to be almost the same as #82. So does this PR replace that one?

mikera:
@kremalicious this change builds on #82 (it's effectively adding an extra asset type on top of that).

The reason for extra structure in `links` is that a Marketplace UI will want to treat different kinds of links differently, so we need a way to specify the link type. Open to ideas on how to do that but this is important if e.g. we want to place sample data in a separate tab.

Happy to do a call if you you want to discuss the metadata changes in more detail?

kremalicious:
Ah sorry if I wasn't clear enough. I was just talking about the style, I don't dispute the `links` structure and I already implemented that in https://github.com/oceanprotocol/pleuston/pull/151. I just don't think it requires its own sub-section, the explanation can simply be left in the table, just like https://github.com/oceanprotocol/OEPs/pull/86 did.

And regarding the example at the end In this PR, you removed that structure and replaced with:

```
"links": [
         {"sample1": "http://data.ceda.ac.uk/badc/ukcp09/data/gridded-land-obs/gridded-land-obs-daily/"},
         {"sample2": "http://data.ceda.ac.uk/badc/ukcp09/data/gridded-land-obs/gridded-land-obs-averages-25km/"},
         {"fieldsDescription": "http://data.ceda.ac.uk/badc/ukcp09/"}
     ],
```

which is not what was already added in https://github.com/oceanprotocol/OEPs/pull/86 and it doesn't follow your own example in this PR:

```
"links": [
            {
                "name": "Sample of Asset Data",
                "type": "sample",
                "url": "https://foo.com/sample.csv"
            },
            {
                "name": "Data Format Definition",
                "type": "format",
                "AssetID": "4d517500da0acb0d65a716f61330969334630363ce4a6a9d39691026ac7908ea"
            }
        ],
```

mikera:
Ah OK - understand now, thanks for the catch!

I think in general this OEP will need further fleshing out, so subsections will probably be necessary going forward. In that spirit, I've converted "additional information" into a sub-section as well.

mikera:
Closing this PR as the relevant content has been moved into the new OEP13.

PR #100, opened 2018-11-27
Edited OEP 8

ttmc:
- Made "Table of Contents" a non-heading so it's not included in the Table of Contents
- Changed all headings to title case
- Regenerated the Table of Contents
- Changed all double-blank lines before headings to single-blank, to conform with standard Markdown
- Rearranged and rewrote the section titled "Assets Metadata Ontology" because it wasn't clear what the the Assets Metadata Ontology _was_ and if the ontology was for the whole DDO or just a part of it. Added a link to OEP 7 which defines the overall structure of an Ocean DDO.
- Lots of minor copy-editing
- Standardized on calling the things "Asset Metadata (objects)", i.e. the things which conform to the Assets Metadata Ontology. (Before, they were sometimes called DDOs, which was totally wrong.)

### DIFF
--- a/8/README.md
+++ b/8/README.md
@@ -25,8 +25,7 @@ contributors: Kiran Karkera <kiran.karkera@dex.sg>, Enrique Ruiz <enrique@oceanp
 
 # Assets Metadata Ontology
 
-Every Asset (dataset, algorithm, etc.) in the Ocean Network has an associated Decentralized Identifier (DID) and DID document / DID Descriptor Object (DDO).
-Why? Assets without proper descriptive metadata can have poor visibility and discoverability.
+Every Asset (dataset, algorithm, etc.) in the Ocean Network has an associated Decentralized Identifier (DID) and DID document / DID Descriptor Object (DDO). Why? Because Assets without proper descriptive metadata have poor visibility and discoverability.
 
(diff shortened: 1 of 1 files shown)


8/README.md
```diff
@@ -7,114 +7,107 @@ editor: Aitor Argomaniz <aitor@oceanprotocol.com>
...
+| **industry**          |                                                                                                                              |
+| **category**          | can be assigned to a category in addition to having tags                                                                     |
+| **updateFrequency**   | how often are updates expected (seldom, annually, quarterly, etc.), or is the resource static (never expected to get updated) |
+| **termsOfService**    |                                                                                                                              |
+| **privacy**           |                                                                                                                              |
+| **keyword**           | A list of keywords/tags describing a dataset                                                                                 |
+| **structured-markup** | A link to machine-readable structured markup (such as ttl/json-ld/rdf) describing the dataset                                |
```
JohnBDB: Tiny thing: we're using "dataset" here, whereas we used "data set" at line 64.

ttmc: I've noticed that we use them interchangeably. I think the reason is that some people like the shortness of "dataset" but unfortunately it's not a real word. However, English is funny in that you can make up words and people will just go with it, and after a while the spellcheckers catch up to reality. "Blockchain" is another example.

JohnBDB: Fair enough. It was more for consistency :)

PR #102, opened 2018-11-28
Add OEP 13 asset metadata draft

mikera:
Draft for OEP 13, based on initial input from OEP 8.

This OEP provides a standardised definition of asset metadata for Ocean Assets, intended to be stored off-chain and accessed via the Meta API.

This is intended as an initial raw OEP for further iteration based on Marketplace / ecosystem tool requirements

### DIFF
--- a/13/README.md
+++ b/13/README.md
@@ -4,7 +4,7 @@ name: Off-chain Asset Metadata
 type: Standard
 status: Raw
 editor: Mike Anderson <mike.anderson@dex.sg>
-contributors: Kiran Karkera <kiran.karkera@dex.sg>, Enrique Ruiz <enrique@oceanprotocol.com>, Mike Anderson <mike.anderson@dex.sg>, Matthias Kretschmann <matthias@oceanprotocol.com>
+contributors: Aitor Argomaniz <aitor@oceanprotocol.com>, Kiran Karkera <kiran.karkera@dex.sg>, Enrique Ruiz <enrique@oceanprotocol.com>, Mike Anderson <mike.anderson@dex.sg>, Matthias Kretschmann <matthias@oceanprotocol.com>
 ```
 
 Table of Contents
--- a/13/README.md
+++ b/13/README.md
@@ -32,9 +32,17 @@ Ocean has a core mission of making data assets visible and discoverable, with a 
 
 As such, each Ocean Asset (dataset, algorithm, etc.) has Asset Metadata associated with it. Asset Metadata is created by the asset publisher, and hashed to ensure integrity and ensure that the asset can be subsequently referenced as part of the provenance of other assets.
 
+Asset Metadata should normally be stored off-chain for the following reasons:
+
+- Eliminate gas cost of requiring on-chain registration of assets
+- Mininise risk of harmful content being published on-chain
+- Keeping metadata private in some trusted data scenarios (e.g. sharing within a consortium) 
+
 Assets without proper descriptive metadata can have poor visibility and discoverability, so it is generally in the publisher's interest to ensure good metadata is made available.
 
-This OEP doesn't detail the exact method of registering or publishing metadata in a metadata store.
+This OEP doesn't detail the exact method of registering or publishing metadata in a metadata store. This is covered in OEP15 (Meta API)
+
+
 
 
 ## Change Process
--- a/13/README.md
+++ b/13/README.md
@@ -1,0 +1,239 @@
+```
+shortname: 13/META
+name: Off-chain Asset Metadata
+type: Standard
+status: Raw
+editor: Mike Anderson <mike.anderson@dex.sg>
+contributors: Kiran Karkera <kiran.karkera@dex.sg>, Enrique Ruiz <enrique@oceanprotocol.com>, Mike Anderson <mike.anderson@dex.sg>, Matthias Kretschmann <matthias@oceanprotocol.com>
+```
+
+Table of Contents
+=================
+
+   * [Table of Contents](#table-of-contents)
+   * [Asset Metadata](#asset-metadata)
+      * [Change Process](#change-process)
+      * [Language](#language)
+      * [Motivation](#motivation)
+      * [Base attributes](#base-attributes)
+        * [Additional Information](#additional-information)
+        * [Links](#links)
+      * [Data asset attributes](#data-asset-attributes)
+      * [Invokable service attributes](#invokable-service-attributes)
+      * [Bundle attributes](#bundle-attributes)
+      * [Example](#example)
+      * [References](#references)
+
+
+
+# Asset Metadata 
+
+Ocean has a core mission of making data assets visible and discoverable, with a decentralised protocol for data exchange.
+
+As such, each Ocean Asset (dataset, algorithm, etc.) has Asset Metadata associated with it. Asset Metadata is created by the asset publisher, and hashed to ensure integrity and ensure that the asset can be subsequently referenced as part of the provenance of other assets.
+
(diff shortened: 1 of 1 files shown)


aaitor:
Hey, question about this PR. It looks a new version of OEP8. Why not sending a PR to OEP8 with the reasoning about the changes? Would be easier to understand what is new or need to be modified.

mikera:
Hi @aaitor this is based on the content from https://github.com/oceanprotocol/OEPs/pull/90 but that didn't seem to be going anywhere.

We need to standardise the Asset metadata for the Marketplace APIs which are in heavy development right now, so I thought best to fork the OEP and put the metadata stuff here (excluding the DDO aspects, which I think are a separate concern and deserve a separate OEP).

If you want to refocus OEP8 on the DDO aspects that would be fine with me. We don't need asset DDOs for the marketplace or tool ecosystem, but if you have another use case for them that is fine of course.

aaitor:
Hi @mikera , I don't understand the comment about OEP-8 doesn't seem to be going anywhere.
OEP-8 standarize the base Metadata of Trilobite release. As been discussed and implemented in Aquarius (reference implementation of Metadata API), Squid (Python, Javascript, Java), used in all the integration tests and parsed in Pleuston. 
That OEP-8 received feedback from different people, and for example some changes you proposed were discussed and merged.

The OEP-8 doesn't include any detail about DDO's. DDO's standarization in Ocean are part of OEP-7 and OEP-11.

Taking into account this OEP-13 is a copy of OEP-8 I don't understand the utility.

mikera:
Hi @aaitor some quick context:
- The marketplace implements its own metadata store - we aren't using Aquarius / OceanDB so what Aquarius does is not particularly relevant.
- We aren't using DDOs for assets and OEP8 says "This metadata, also called DID descriptor objects (DDO)" so OEP8 seems incompatible with our approach (at least as currently defined)
- We need to make various design changes to the asset metadata to support various marketplace and ecosystem functionality (e.g. provenance). 

Previous discussions on DIDs, DDOs and Asset Metadata appear to have been unproductive as we haven't got consensus on key design decisions, so I don't think it would be helpful to go down the route of sending more PRs on OEP8. 

It therefore seems simpler if we fork and maintain this OEP separately right now. Unless you have an better suggestion? Want to do a quick call to discuss?

diminator:
I would disapprove this to be called a `standard`. It looks more like a marketplace convenience API that is not required for each marketplace to implement.

shark8me:
For comparison's sake, the ethereum json rpc api doesn't call itself a
standard either, it is defined in the wiki and not in an EIP.
Wikis don't work well with version control however, and so the API is
better defined as a version controlled document. Are you suggesting that
layer 2 and 3 APIs are not called OEPs?

mikera:
@diminator - this isn't an API, it's the standard for asset metadata. We've determined that this needs to be standardised to enable interoperability between marketplaces and service providers. 

The relevant API for accessing this will be described in OEP15 (PR pending)

diminator:
> Are you suggesting that layer 2 and 3 APIs are not called OEPs?

- Layer 1 contains the smart contracts and the associated state. It's the only source of truth given adversarial assumptions (Byzantine, Sybil). Security guarantees such as immutability, integrity and soundness are strong at this layer. Marketplace transactions that want to flow through this layer need to comply with the specifications at Layer 1. Hence layer 1 defines standards that span across the entire network.
- Layer 2 contains the middleware to expose the APIs of Layer 2 into multiple environments (JS, python, Java, ...). The interface between Layer 1 is defined by those standards. This layer implements abstraction and orchestration. The latter (orchestration) should be uniform across all implementation (standard). There is a minimal set of functionality that Layer 2 needs to expose (standard). And a set of functionality that can be regarded as utilities (convenience).
- Layer 3 contains the services such as resources, assets, marketplaces etc. Different verticals will have different domain standards. I think we should we careful here in what to standardize, as this could conflict with what the local community, clan, tribe, group, marketplace regards as standards and introduces onboarding friction as such

diminator:
> this isn't an API, it's the standard for asset metadata. We've determined that this needs to be standardised to enable interoperability between marketplaces and service providers.
> 
> The relevant API for accessing this will be described in OEP15 (PR pending)

Metadata is a service (Layer 3). Layer 1 and 2 care about the identifiers (DID) and how to resolve them. 
One question we could ask ourselves

> We've determined that this needs to be standardised to enable interoperability between marketplaces and service providers.

[OEP-8](https://github.com/oceanprotocol/OEPs/tree/master/8) address the interoperability issue through the introduction of DDOs and DIDs. This is the contract between Layer 1 and Layer 2 in order to successfully resolve a Service Agreement. Given that you are proposing another standard for such resolution, this would require to be provisioned in Layer 2.

diminator:
> We need to standardise the Asset metadata for the Marketplace APIs which are in heavy development right now, so I thought best to fork the OEP and put the metadata stuff here (excluding the DDO aspects, which I think are a separate concern and deserve a separate OEP).

Layer 1 contracts require DIDs. If you don't want to use DIDs - go ahead and fork the contracts. That's another project - not Ocean Protocol

mikera:
@diminator I think we need to resolve a fundamental difference in perspective about what an asset represents, and how it is addressed. From the W3C DID spec (https://w3c-ccg.github.io/did-spec/#paths):

`A DID path SHOULD be used to address resources available via a DID service endpoint`

We view assets as resources that may be served up by potentially many service providers, and as such should be referenced using a DID path such as:

`did:op:f8d3e155d53735fc1cbf5418500ac32c1dfeb1157bed7e9d47978ff51dc4ca2e/eb9fe47d8b420b6ca3fc798c523fe5359b9a08d283a63c21ce1559ec1ff1d350`

Where the first hex part before the `/` is the DID of the service provider (which needs a corresponding DDO) and the second part is the asset ID (which doesn't require a DDO, it's just a resource). The asset ID is simply a hash of the OEP13 metadata, which makes it useful to verify the integrity of metadata as well as the signature of whoever includes it in an on-chain transaction.

As I understand your current position, you are conflating the service provider with the asset to create a single DID/DDO for the asset which includes service provider endpoints. I see that as problematic for multiple reasons, chiefly:

- It means that every asset will require a DID/DDO registered on chain, which is expensive (gas+network bandwidth) and will be a barrier to adoption (imagine registering 100,000 open data assets, for example...). The DID path solution only requires service providers to register a DID/DDO, which will be substantially smaller in number.
- It makes it complicated to handle cases where the same asset is served up by many service providers (many use cases for this: mirroring/caching, sharing of assets which represent open source data/algorithms, cross-indexing of assets on different marketplaces etc.). With DID paths this is trivial: you just use different service provider DID with the same asset ID.
- It is IMHO inconsistent with the W3C did spec as referenced above, which states that resources should be referenced by did paths, not via independent DIDs

If I may draw an analogy, the DID is like a host name, the asset is like a URL path to a specific resource on a website. It would be a strange and highly inconvenient design decision for the web if you had to register every individual html file with a DNS record, but that appears to be what you are proposing in the Ocean space? Please correct me if I have misunderstood your position of course.

diminator:
What matters at this point is that the asset ID (or path) is resolvable.

Assets won't have a single endpoint - they will have a more complex structure to interact with them. For example:
- how will access be granted (multiple endpoints)
- what service agreement template does the asset connect to
- what are the events to listen and react to (multiple endpoints)
- the actual metadata describing the asset

So it's not "just an URL path", but an entire flow.

I think both approaches could co-exist (DID and DID path/embedding the asset in the provider) but they would require different resolvers and take into account the requirements for each asset to connect to Layer 1. Some potential issues that could arrise:
- co-ownership and liquid governance of assets (ie there's not a single provider)
- updates of assets require updates of the provider
- lengths of the DID/path need to fit the contract bytesize

PR #106, opened 2018-11-30
Feature/oep12 provenance

shark8me:
The first draft of provenance metadata to be generated on publishing new assets, importing assets, or derivations of existing assets.

requesting review from @mikera  and @oceanprotocol/maintainers.


kremalicious:
was merged really quickly before work day starts in Europe, so here's my review:

- needs the standard block at the top as in all other OEPs, with it marking clearly the status of this OEP. No-one from the outside is able to figure out if this is raw, draft, or final
- needs a proper TOC as in all other OEPs
- why so many quote styles in there? Are these all quotes? Then this should be said above or below each quote
- If merged to `master`, the OEP should be included in the readme list, which this PR hasn't done, missing another chance telling outside people the status of this OEP
- Not super important but I would dial back on the usage of all those bullet and nested bullet lists, this makes the OEP hard to read

PR #108, opened 2018-12-01
Fix OEP README - placeholders and numeric assignments

mikera:
This PR re-establishes placeholders for marketplace and ecosystem API OEPs previously removed in error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,12 +33,12 @@ Short Name          | Title                                                     
 [3/ARCH](3)         | Ocean Core Architecture                                      | Standard | Draft        | Dimitri De Jonghe
 [4/AGENT](4)        | Ocean Agent Protocol Stack                                   | Standard | Draft        | Aitor Argomaniz
 [6/INVOKE](4)       | Invokable Compute Services                                   | Standard | Raw          | Kiran Karkera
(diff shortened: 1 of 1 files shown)


kremalicious:
Adding dead links can hardly be called a "fix". Setting status to "To be merged" creates the expectation there're open PRs for those OEP but they are nowhere to be found. Likewise, I am not able to find that status in the [COSS lifecycle](https://github.com/oceanprotocol/OEPs/tree/master/2#coss-lifecycle).

So in this format this just adds confusion cause to users, these are simply not placeholders, and status doesn't follow the patterns we defined in OEP-02. If you want placeholders for non-existent work, it should be properly labeled like that, add a new table "Upcoming OEPs" or whatever, but then again the list of PRs should be all that's needed to show work on upcoming OEPs

diminator (changes requested):
This was not removed in error.
It was removed because of not following community guidelines as specified in [2/COSS](https://github.com/oceanprotocol/OEPs/tree/master/2)

There is no need to cause confusion by:
- adding empty or dead links
- incomplete drafts without consensus
- proposals for non-standards

kremalicious:

PR #111, opened 2018-12-04
Proposals have not been reviewed 

diminator:
Both proposals have not been reviewed from:
- a protocol architecture view
- if they are relevant protocol standards or API guidelines

1) CSAPI: unfinished draft, not sure about relevance
2) Provenance: based off metadata, not on-chain verified proofs

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,13 +32,11 @@ Short Name       | Title                                                        
 [2/COSS](2)      | Consensus-Oriented Specification System                      | Meta         | Draft      | Dimitri De Jonghe
 [3/ARCH](3)      | Ocean Core Architecture                                      | Standard     | Draft      | Dimitri De Jonghe
 [4/AGENT](4)     | Ocean Agent Protocol Stack                                   | Standard     | Draft      | Aitor Argomaniz
-[6/CSAPI](6)     | API to register & invoke computer services                   | Standard     | Raw        | Mike Anderson
 [7/DID](7)       | Decentralized Identifiers									| Standard     | Raw        | Aitor Argomaniz
 [8/ASSET-DDO](8) | Assets Metadata Ontology                                     | Standard     | Raw        | Aitor Argomaniz
 [9/BOUNTY](9)  | Bounties in Ocean Protocol                       	 		 | Standard | Raw      | Chalid Mannaa, Manan Patel
 [10/OAA](10)     | On-Chain Access Control                       				| Standard     | Draft      | Ahmed Ali
 [11/ACL](11)    | On-Chain Access Control using Service Agreements           | Standard | Raw      | Aitor Argomaniz
-[12/PROV](12)    | Provenance metadata | Standard | Raw      | Kiran Karkera 
 
 
 


mikera (changes requested):
@diminator I have reviewed these and determined that they are appropriate to be included as Raw OEPs at this stage, for further development.

You are welcome to propose amendments or enhancements to specific areas of content of course.

kremalicious (approved):
OEP-12 was merged before anyone outside of Singapore timezone was able to review anything, and it wasn't even following the [basic formatting standards](https://github.com/oceanprotocol/OEPs/tree/master/2#oep-template) we defined in OEP-02. How can that be considered something to be merged to `master` if it violates our previously agreed upon standards?

And existence of this PR here shows that there is no "rough consensus" for these OEPs, to me it looks like the existence of those OEPs is something which needs to be discussed, not some amendment. So we should remove them from the list until that's done.

diminator:
> @diminator I have reviewed these and determined that they are appropriate to be included as Raw OEPs at this stage, for further development.

1) Empty placeholders do not qualify as raw [CSAPI](https://github.com/oceanprotocol/OEPs/tree/master/6)
2) non-standards (ie non-mandatory to be implemented by > level 1) might want to have a different home. For example a wiki, or an API documentation. I would say [PROV](https://github.com/oceanprotocol/OEPs/tree/master/12) is a convenience API. While something like verified provenance (ie smart contract state, cryptographic proofs, etc.) would qualify as a standard. Metadata provenance is useful, but not enforced, has no required integrity, etc. Unless a project pops up with a verified, decentralized metadata store (ie. BigchainDB-style).

shark8me:
@diminator 

OEP-12 is a suggested provenance metadata standard for tools that need provenance information attached, especially for services that generate datasets from other assets (e.g. join on 2 datasets produces a new dataset). Yes, It is not using on-chain verified proofs. Are you suggesting that provenance is only acceptable if based on proofs?

diminator:
> Are you suggesting that provenance is only acceptable if based on proofs?

I think there's not much value in off-chain provenance in an adversarial network. At least not for dispute resolution. The proposed standard could work with integrating on-chain provenance. Or maybe referring to the source of provenance? This would allow dispute resolution to assess the quality of provenance claims

mikera:
@diminator I believe a full story for provenance requires the combination of two things:
a) A standard for publisher-declared provenance metadata as part of an asset. OEP12 addresses this. This is useful for ecosystem / tool integration, and is reliable to the extent that consumers are willing to trust whoever signed the asset metadata, which will be true in many cases.
b) A standard for on-chain information that enables actors to verify what happened to generate an asset (e.g. a service invocation). This can and probably should be used to verify some aspects of b) where possible, which I guess is what you mean by integrating with on-chain provenance for dispute resolution?

Basically, the intent is to have the on-chain and off-chain aspects of provenance work together to give asset consumers a more complete (and where possible, verifiable) picture of asset provenance. 

Verification can be addressed in UI, e.g. the marketplace client asking squid to confirm on-chain provenance and potentially giving a visible warning if something cannot be verified. Verification can also be done at the marketplace backend, e.g. a marketplace refusing to list assets when publisher declared provenance contradicts what is observable on-chain.

Suggest we have a call if you want to discuss in more detail?

mikera:
@kremalicious - as per accountability agreed at the Ocean Protocol board, OEPs relating to the marketplace / tool ecosystem such as these are being managed by Singapore, in exactly the same way that OEPs related to the network are being managed by Berlin. 

You are very welcome to comment and suggest improvements to the content of course, but from a process perspective these additions do not require review or approval from Berlin. I suggest you discuss with Bruce if you require further clarification.

diminator:
> Basically, the intent is to have the on-chain and off-chain aspects of provenance work together to give asset consumers a more complete (and where possible, verifiable) picture of asset provenance.

That sounds reasonable. Maybe this flow can be explained in the proposal?
In the end this will require the source of provenance to be listed in the provenance objects.
JSON-LD might be a tool there.
I would like to avoid confusion when people see provenance but cannot figure out how it's verified.
Maybe there are existing provenance tools that already solve what you suggest but just need an on-chain extension?

> Verification can also be done at the marketplace backend, e.g. a marketplace refusing to list assets when publisher declared provenance contradicts what is observable on-chain.

I don't see how "verification" can be done as such? We're looking at verification as cryptographic proofs, attestations, hashes, signatures, etc. Verified provenance is untrusted provenance with a proof layer.

diminator:
also, this is not the right place for the provenance discussion, #112 is

kremalicious:

PR #112, opened 2018-12-06
Examples for oep12 provenance

shark8me:
This PR adds complete examples of provenance sections  included in the asset metadata for the use cases of

* initial publishing
* services creating derived assets


diminator (changes requested):
I think the big question is "How can the provenance be trusted?"

- there are no descriptions on how the provenance is established
- how is integrity and soundness addressed?
- does this have cryptographic support that is verified at level1?

diminator:
related discussion https://github.com/oceanprotocol/OEPs/pull/111

mikera:
@diminator see my comments on #111 but from a design perspective I certainly agree that that we should be verifying provenance against service agreements wherever possible. 

Perhaps you can suggest a concrete way that the service agreement should be referenced in the provenance format to make this possible? We would obviously need corresponding functionality in the squid API to preform the actual verification.

From a collaborative process perspective, please either approve the PR or suggest an actionable improvement that @shark8me can incorporate. I would prefer to merge fast and iterate rather than having PRs stuck in limbo without an obvious next step.

12/examples/cleaned_dataset_metadata.json
```diff
@@ -0,0 +1,78 @@
...
+              },
+              "ocn:did": "did:ocn:anotherbigname3572"
+          }
+      },
+      "activity": {
+          "ocn:ocn_asf29fj291lkd": {
+              "prov:label": "Cleaning the dataset",
+              "prov:endTime": "2018-02-02T14:07:13",
+              "prov:startTime": "2018-02-02T11:03:23",
+              "prov:type": "ocn:service"
```
mikera: I believe this section should include a reference to the on-chain service agreement used to create the asset, if it exists?

12/examples/cleaned_dataset_metadata.json
```diff
@@ -0,0 +1,78 @@
...
+          "ocn": "http://www.oceanprotocol.com/prov"
+      },
+      "agent": {
+          "ocn:ocn_abvfwi89sjek": {
+              "prov:label": "Hospital data administrator",
+              "prov:type": {
+                  "$": "prov:Person",
+                  "type": "prov:QUALIFIED_NAME"
+              },
+              "ocn:did": "did:ocn:llongameabcd2938"
```
mikera: What does this did refer to? The individual administrator? If so this needs to be specified in the main OEP text, and we need to confirm if there is an design for individuals (as opposed to Ocean Agents) have a on-chain-registered DID.  

Suggest this is optional (we don't want to make it mandatory for individual users of Ocean to register DIDs), but if so that needs to be clearly specified.

12/examples/cleaned_dataset_metadata.json
```diff
@@ -0,0 +1,78 @@
...
+          },
+          "ocn:ocn_abv2837282isk": {
```
mikera: Is this identifier simply an arbitrary ID valid only for the scope of this provenance  file? Should be specified how this behaves.

mikera (changes requested):
Various comments to address. In particular, I think there needs to be a bit more formal definition / clarification what the different fields mean.

12/examples/cleaned_dataset_metadata.json
shark8me: Agreed, will add that.

12/examples/cleaned_dataset_metadata.json
```diff
@@ -0,0 +1,78 @@
...
+          "ocn:ocn_asf29fj291lkd": {
+              "prov:label": "Cleaning the dataset",
+              "prov:endTime": "2018-02-02T14:07:13",
+              "prov:startTime": "2018-02-02T11:03:23",
+              "prov:type": "ocn:service"
+          }
+      },
+      "entity": {
+          "ocn:this": {
+              "prov:value": {
```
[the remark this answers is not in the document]
shark8me: I'm inclined to have this as mandatory, so that the raw prov section should be interpretable directly rather that having to reference the metadata.

12/examples/cleaned_dataset_metadata.json
```diff
@@ -0,0 +1,78 @@
...
+                  "$": "A dataset of cleaned parkinsons sensor readings",
+                  "type": "xsd:string"
+              },
+              "prov:type":"ocn:dataset",
```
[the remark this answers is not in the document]
shark8me: Again, it is useful from a PoV of the prov section being self-contained. The options will be defined in the validation tooling in a separate repo.

12/examples/cleaned_dataset_metadata.json
shark8me: This is the user account with an "ocn_" prefix, could be web3 or marketplace defined (length). Will be specified in the validation tooling.

I should  clarify that these examples are not mean to replace a proper spec & validation library.

shark8me:
@mikera I've added my comments. The formal spec & validation tooling, which specifies the mandatory/optional and type enums, will be a separate repo.

kremalicious:
Seems abandoned, requested changes were not provided, so closing.

mikera:
Provenance has moved here - https://github.com/DEX-Company/DEPs/tree/master/12

PR #123, opened 2019-01-11
WIP - Clarification in RSA encryption (OEP11)

aaitor:
Don't merge, some clarifications are pending


ttmc:
What should be done with this pull request? Is it defunct so it can be closed? Should it be updated?

PR #126, opened 2019-01-16
Feature/oep13 commons marketplace

aaitor:


13/README.md
```diff
@@ -0,0 +1,275 @@
...
+1. The CONSUMER select one of the ASSETS and open the **detailed view page** using the MKT app
+
+1. The MKT app draw in the **detailed view page** all the metadata associated to the asset. The asset URL must be unique, and include the DID in the URL. Example: 
+   `http://commons.oceanprotocol.com/asset/did:op:4d517500da0acb0d65a716f61330969334630363ce4a6a9d39691026ac7908ea`
+
+   Also should include all the metadata tagged in the page allowing an easy SEO for the contents.
+
+1. The **detailed view page** shows a button or download link for each of the content urls existing in the DDO. 
+ 
+1. When the CONSUMER clicks on the "DOWNLOAD" link of any of the contents, the CONSUMER using SQUID-JS initialize the purchase process.
```
kremalicious: it's not clear why there should be a purchase and download flow for the commons marketplace. Most links on Google data set search point to websites from where the files can be obtained, not the files itself. So adding a purchase flow seems a bit unnecessary here.

13/README.md
```diff
@@ -0,0 +1,275 @@
...
+   - List of public URL's where the Asset is located
+   - Title
+   - Description
+   - Category (categories can be predefined upfront)
+   - Author
+   - Tags
+   
+   A complete list of attributes can be found in the [OEP-8](https://github.com/oceanprotocol/OEPs/tree/master/8).
+   
+1. The MKT validate that URL's are correct and contents are available in those URL's initiating the download of the contents. Contents are not persisted in any place, the MKT application only validate that content exist in the URL's provided. If any URL is invalid, the processed is cancelled and the user is informed of the error.
```
kremalicious: what's "initiating the download of the contents" when a url to a website is provided? That's the default on Google data set search

13/README.md
```diff
@@ -0,0 +1,275 @@
...
+
+1. The MKT validate the input fields provided by the user. If information is missing, the process is cancelled and the user is informed of the error.
```
kremalicious: nothing is cancelled, we have normal HTML input validation going on so form can't be submitted when required fields are missing

13/README.md
```diff
@@ -0,0 +1,275 @@
...
+
+1. The PUBLISHER, search on [Google Datasets](https://toolbox.google.com/datasetsearch/) website  and discover a relevant free/commons asset
+
+1. The PUBLISHER goes to the Ocean Commons Marketplace frontend application, and open the form of "Import a Commons/Free Asset from Google Datasets"
+
+1. The PUBLISHER fill the following information:
+   - URL of the Google Dataset to import (i.e: https://toolbox.google.com/datasetsearch/search?query=Weather%20in%20Australia&docid=gEf986q6CT26lj9yAAAAAA%3D%3D)
+   - List of URL's of the contents associated (can't be imported directly)
+
+1. The MKT app retrieve the metadata from the Google Dataset URL
```
kremalicious: this is still a big question mark. As it looks like Google does not markup their search results with proper schema.org metadata in the `head`. They also don't have an API. Their page markup is not semantic with auto-generated class names. Unless at least one of those things changes, I doubt we can implement this.

13/README.md
aaitor: but it's what we do in Ocean, we put in the same place assets (in this case free/commons) and allow the discovery, consumption of those.

13/README.md
```diff
@@ -0,0 +1,275 @@
...
+          }
+        ]
+ ```
+
+For the network rewards, the Smart Contracts need to have the internal logic to calculate the rewards in free and paid scenarios.
+In the free scenarios, the Smart Contracts will check if prize is zero, in that case the
+
+In the `releasePayment` method, it's necessary to validate that parameters are not having a higher value than the caps defined by the system.
+
+### User Interface Screens
```
kremalicious: please don't add this here. The plan for this is in development right now and differs in various stages, first prototype, MVP, etc. https://github.com/oceanprotocol/commons-marketplace/issues/2

13/README.md
aaitor: We shouldn't provide URL's. We should allow downloading the asset's binary.

13/README.md
aaitor: I was trying to refer to the normal procedure of execution, where the input validation is the first phase. But I guess can be rephrased to give more clarity.

13/README.md
aaitor: Yep, maybe all the "import" functionality could be a potential MVP-X pending of a previous research

13/README.md
aaitor: OEP should work as architecture definition and is coming before development. So high-levels of the flow we want to achieve should be here, implementation details in the specific repos.

13/README.md
```diff
@@ -0,0 +1,275 @@
...
+
+
+-----
+
+
+# Ocean Commons Marketplace use case
+
+The intention of this document is to describe the scenario of a working Ocean Commons Marketplace. There the users should be able to publish and discover free/commons assets (datasets). 
+
+It's out of the scope to detail the detailed implementation of components (can be found in the [OEP-11](https://github.com/oceanprotocol/OEPs/tree/master/11)
```
ssallam: `It's out of the scope to detail the implementation of ...`

13/README.md
```diff
@@ -0,0 +1,275 @@
...
+            }, {
+             "name": "price",
+             "type": "uint",
+             "value": 0
+          }
+        ]
+ ```
+
+For the network rewards, the Smart Contracts need to have the internal logic to calculate the rewards in free and paid scenarios.
+In the free scenarios, the Smart Contracts will check if prize is zero, in that case the
```
ssallam: This sentence seems incomplete.

13/README.md
```diff
@@ -0,0 +1,275 @@
...
+
+1. BRIZO listens for the `PaymentLocked` event and when it's received call the `GrantAccess` function in the KEEPER
+
+1. The Keeper emit the `AccessGranted` event
+
+1. The CONSUMER listens for the `AccessGranted` event and when it's received call BRIZO sending the encrypted URL
+
+1. BRIZO valide on-chain using the `checkPermissions` function in the KEEPER Access Conditions if the user has access. If not cancel the request returning a `HTTP 401 Unathorized` answer.
+
+1. BRIZO, after validate the permissions decrypt the encrypted URL using the PROVIDER Keys
```
ssallam: Brizo does not actually decrypt anything, the consumer decrypts the urls before invoking Brizo's consume endpoint.

13/README.md
```diff
@@ -0,0 +1,275 @@
...
+
+1. BRIZO, working as a PROXY, return the Asset content to the CONSUMER
+
+The complete PUBLISH and CONSUME flows can be shown in the below image:
+
```
ssallam: The details of what Brizo does seem unnecessary here. The MKT only needs to send the request, lock the payment, listen to AccessGranted event, then download the datasets.

13/README.md
```diff
@@ -0,0 +1,275 @@
...
+ 
+1. When the CONSUMER clicks on the "DOWNLOAD" link of any of the contents, the CONSUMER using SQUID-JS initialize the purchase process.
+
+1. SQUID-JS calls to BRIZO to initialize the Service Executing Agreement (SEA)
+
+1. BRIZO initialize the SEA on-chain using the KEEPER
+
+1. The KEEPER after to initialize emit the `ExecuteAgreement` event
+
+1. The CONSUMER listens for the `ExecuteAgreement` event and when it's received call the `LockPayment` function on the KEEPER
```
ssallam: Keeper functions start with lowercase, events are correct though.

13/README.md
aaitor: fixed

13/README.md
aaitor: fixed

13/README.md
aaitor: yep, but I think the details are important to understand what need to be implemented from the backend side

kremalicious:
Closing in favor of https://github.com/oceanprotocol/OEPs/pull/200 which is exact copy of this PR's content. Rebasing this one was not fun.

PR #128, opened 2019-01-18
Reconciliation of OEP11 with new files attribute of OEP8 instead of contentUrls

aaitor:


11/README.md
```diff
@@ -447,19 +447,34 @@ Example:
...
+   ]
 ```
 
 In this case, the following text will be encrypted:
 ```
-["https://example.com/file1.csv","https://example.com/file2.csv"]
+[{"url":"234ab87234acbd09543085340abffh21983ddhiiee982143827423421","checksum":"efb2c764274b745f5fc37f97c6b0e761","contentLength":"4535431","resourceId":"access-log2018-02-13-15-17-29-18386C502CAEA932" }, { "url":"234ab87234acbd6894237582309543085340abffh21983ddhiiee982143827423421","checksum":"085340abffh21495345af97c6b0e761","contentLength":"12324"},{"url":"80684089027358963495379879a543085340abffh21983ddhiiee982143827abcc2"}]
 ```
 
 After the encryption, the previous URL's will be removed and the encrypted HASH added to the DDO.
```
JohnBDB: URLs instead of URL's

PR #139, opened 2019-02-04
Authorization as a service (secret store)

aaitor:
Moving secretStore url to an independent service. More clear, generic and did spec compliant


Pedro-vk:
If the user uses `None` as Authorization service, files if going to be sent without encryption and `encryptedFiles` as an empty array?

aaitor:
If the DDO doesn't specify the secret store, the default given in the squid libraries will be used (by default the Ocean secret store cluster)

Issue #140, opened 2019-02-04
Update OEP8, various


MarcusJones:
`compression` field should be moved into`files` array, since files can have different compression
`contentType` field should be moved into`files` array, since files can have different types
`encoding` field should be moved into`files` array, since files can have different types

`links` spec is not listed in readme, but does appear in example;
```
        "links": [
            {
                "name" : "Sample of Asset Data",
                "type" : "sample",
                "url": "https://foo.com/sample.csv"
            }
        ]
```

MarcusJones:
~~isDeleted~~ -> isListed

ttmc:
Pull request #154 (now merged) moved the file-related metadata inside the `files` array, to that is done.

The current OEP-8 spec _does_ mention the `links` array as part of the spec, and its description says, "Mapping of links for data samples, or links to find out more information. Links may be to either a URL or another Asset. We expect marketplaces to converge on agreements of typical formats for linked data: The Ocean Protocol itself does not mandate any specific formats as these requirements are likely to be domain-specific."

I interpret that to mean that the exact structure of the `links` array members is free-form, and if there is any consistency in the examples, those shouldn't be interpreted as any kind of _de facto_ spec for what the members should look like (other than being "links" in some sense).

Therefore, in my opinion, this issue can be closed.

PR #154, opened 2019-03-26
OEP-8: move file-related metadata to files array

kremalicious:
Putting all those file-related metadata attributes to files array makes the metadata model more flexible so it can be used no matter how many files are added for an asset.

### Current state

When more than 1 file is added to an asset, the following attributes under `base` don't make sense anymore since they depend on the added files and they can be different for each file:

- `contentType`
- `compression`
- `encoding`
- `size`

### Proposal

Move all those attributes under the respective entry in the `files` array. We can kick out `size` since there's already `contentLength` present for each file. Resulting in:

```json
"files": [
    {
        "contentType": "application/zip",
        "encoding": "UTF-8",
        "compression": "zip",
        "checksum": "2bf9d229d110d1976cdf85e9f3256c7f",
        "checksumType": "MD5",
        "contentLength": 12057507,
        "url": "https://s3.amazonaws.com/assets/training.zip"
    },
    {
        "contentType": "text/txt",
        "encoding": "UTF-8",
        "compression": "none",
        "checksum": "354d19c0733c47ef3a6cce5b633116b0",
        "checksumType": "MD5",
        "contentLength": 928,
        "url": "https://s3.amazonaws.com/datacommons/monkey_labels.txt"
    },
    {
        "contentType": "application/zip",
        "url": "https://s3.amazonaws.com/datacommons/validation.zip"
    }
],
```


ttmc:
This pull request would resolve some (but not all) of the issues raised in #140

aaitor:
Looks interesting, we can discuss after tethys release

PR #155, opened 2019-03-27
documentation of the flow supporting encryption/decryption in metamas…

aaitor:
…k scenarios

### DIFF
--- a/11/README.md
+++ b/11/README.md
@@ -11,6 +11,7 @@ contributors: Lev Berman <ldmberman@gmail.com>,
 
 ```
 
+
 Table of Contents
 =================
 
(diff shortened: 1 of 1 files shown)


11/README.md
```diff
@@ -448,6 +456,30 @@ Using those parameters, BRIZO does the following things:
 
 ![Consuming Flow](images/consuming-flow.png)
 
+#### Consuming without direct integration of Secret Store
+
+If the CONSUMER (via Squid) can't integrate directly Secret Store for decryption (Squid-js using Metamask can't provide the account password), 
+it's possible to call Brizo with an alternative `consume` method.
+
+In this scenario, it's BRIZO as provider the one in charge of decrypting the content in behalf of the CONSUMER.
```
ttmc: Using the word "provider" here is confusing, as least for me.

We used to use the term "Provider" for the software which was refactored into Aquarius and Brizo. It was my impression we were trying to avoid it since then.

It seems this document redefines the word "provider" to mean some service (such as a Secret Store) which does "URL decryption and SEA management". Let's use a more specific name that "providers" for those actors. How about "Crypto-providers" or "CRYPTO-PROVIDERS", i.e. another actor similar to PUBLISHERS and CONSUMERS? (What does "SEA management" even mean?)

P.S. While searching this document for the word "provider", I noticed an old endpoint `http://myaquarius.org/api/v1/provider/assets/metadata/{did}` (above) which is out of date. The substring `/provider/` should be replaced by `/aquarius/`

PR #162, opened 2019-04-11
OEP12 Draft: Execution of Computing Services

aaitor:


12/README.md
```diff
@@ -0,0 +1,388 @@
...
+A workflow consists of an orchestrated and repeatable pattern of activities transformed into processes that transform or process information.
+
+In Ocean we use the concept of workflow to represent a list of tasks to accomplish with the intention of process data.
+
+From a technical point of view, a workflow is a type of Asset (it takes advantage of all the Ocean plumbing of registering, metadata publishing, resolving, etc.). 
+The main objective of a workflow is to describe an execution pipeline. A workflow can be splitted in sequential stages, having each stage input, algorithm and output.
+
+In the below example, a workflow is modeled in a JSON document with the following characteristics:
+
+* It has a list of sorted stages by the `stages.index` parameter to be executed sequencially
```
ttmc: I wonder about workflows where some output coming from stage A becomes an input for both stage B and stage C, where stage B and C can run at the same time (in parallel).

12/README.md
```diff
@@ -0,0 +1,388 @@
...
+* Each stage has a list of optional minimum requirements, like the image where it will be executed, cpu required, memory, etc.
+* Each stage has an array of sorted input parameters. Each input parameter may be: 
+  - A DID (example: `"id": "did:op:12345"`)
+  - The output of a previous stage (example: `"previousStage": 0`)
```
ttmc: I wonder about stages which can take inputs from several previous stages. Maybe you chose to keep workflows as a simple sequence of stages, with the intent to generalize later? If so, there should be some comment about that.

ttmc: The example below seems to have a stage that takes two inputs...

12/README.md
```diff
@@ -0,0 +1,388 @@
...
+         * [Setting up the Service Execution Agreement](#setting-up-the-service-execution-agreement)
+         * [Execution phase](#execution-phase)
+      * [Infrastructure Orchestration](#infrastructure-orchestration)
+
+
+
+---
+
+
+# Execution of Computing Services using Service Agreements
```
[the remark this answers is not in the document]
aaitor: changed

12/README.md
```diff
@@ -0,0 +1,388 @@
...
+* Detail the main characteristics of this interaction
+* Specify the pros and cons of this approach
+* Identify the modifications required to integrate in the Ocean stack
+* Identify the API methods exposed via the different libraries
+
+## Actors
+
+The different actors interacting in this flow are:
+
+* PROVIDERS - Provide access to the Computing Services
```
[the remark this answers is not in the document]
aaitor: rephrased

12/README.md
aaitor: The reference from a stage to a previous stage output can be achieved using the `previousIndexStage`

12/README.md
aaitor: The current definition doesn't support stages running in parallel having dependencies of other stages. I would leave this for a more advanced scenario.

ttmc:
Maybe you could add a comment to note that the spec currently only supports serial workflows (i.e. a stage can have 0 or 1 previous stages, and 0 or 1 next stages).

PR #167, opened 2019-05-23
In OEP-8, change price from number to string

ttmc:
This pull request is to help with the issue https://github.com/oceanprotocol/ocean/issues/301

- Change "price" from a number/integer to a string (i.e. an integer inside a string).
- Make it clear that the price is in vodka (attoOCEAN), not OCEAN.
- Since the price is required, there is no default value, so I removed the mention of a default value.
- Updated the two price examples.


ttmc:
Now that I think about it, the example prices ("230" and "10" vodka) are probably unrealistically low. I'll make them bigger.

kremalicious:
merging this cause we are already in full swing updating components with this change

PR #169, opened 2019-05-23
Update & make fixes to OEP-8

ttmc:
- Clarified the meanings of "local metadata" and "remote metadata". Noted that both the publisher's Squid, _and Aquarius_, might both make changes.
- "base", "curation" and "additionalInformation" should all be on the same level of the hierarchy, so I fixed that by adding a new **Metadata Attributes** subsection. Also removed "curation" and "additionalInformation" from the list of base attributes.
- The overall "checksum" is only required in the remote metadata. The publisher doesn't know it until they publish, because it depends on the DID.
- Completely replaced the two examples (local and remote metadata), based on this actual DDO in Commons Marketplace: https://aquarius.marketplace.dev-ocean.com/api/v1/aquarius/assets/ddo/did:op:47cd2fa549594bd09e7c024dd5b8252b407fc19cc3f344b8b5b2a4103a4ddc25 
- Revised some heading levels and regenerated the Table of Contents.
- Added myself to the list of contributors, which I also reformatted for readability.


8/README.md
```diff
@@ -138,162 +159,85 @@ These are examples of attributes that can enhance the discoverability of a resou
...
-        "checksumType": "MD5",
-        "contentLength": 12057507,
-        "url": "https://s3.amazonaws.com/assets/training.zip"
-      },
-      {
-        "index":1,
-        "contentType": "text/txt",
-        "encoding": "UTF-8",
+        "url": "https://example-url.net/weather/forecast/madrid/350750305731.xml",
+        "found": true,
```
kremalicious: we can remove `found`, it is only something the file checker in the commons marketplace does for presentation in commons. We send that attribute upon registering but actually shouldn't

8/README.md
```diff
@@ -138,162 +159,85 @@ These are examples of attributes that can enhance the discoverability of a resou
...
-      },
-      {
-        "url": "https://s3.amazonaws.com/datacommons/links/sample/sample.zip",
-        "name": "sample.zip",
-        "type": "sample"
-      }
-    ],
-    "inLanguage": "en"
+    "links": [],
+    "tags": "",
```
kremalicious:
```suggestion
    "tags": [],
```

8/README.md
```diff
@@ -138,162 +159,85 @@ These are examples of attributes that can enhance the discoverability of a resou
...
     ],
-    "inLanguage": "en",
-    "curation": {
-        "rating": 0.93,
-        "numVotes": 123,
-        "schema": "Binary Voting",
-        "isListed": true
-     }
+    "links": [],
+    "tags": "",
```
kremalicious:
```suggestion
    "tags": [],
```

8/README.md
```diff
@@ -58,21 +68,34 @@ The main motivations of this OEP are to:
...
+### Local Metadata
 
-As such, there are 2 flavors of metadata:
+Metadata is first created by the publisher of the asset. The publisher has knowledge of the file URLs, and they are stored in plaintext in the `files` object. This initial metadata is the "local metadata."
 
-1) Local metadata - Created by the publisher of the asset, added to the `DDO` and sent to the metadata store API. 
+### Remote Metadata
 
-2) Remote metadata - Created and existing internally by the metadata store, and sent as a result of querying the API for discoverability of assets. 
+A publisher runs Squid on their local machine. When they publish (register) an asset, they give their local metadata to their local Squid. Squid makes some changes and additions in the metadata, puts it into a DDO, and sends that DDO to a metadata store (Aquarius).
```
kremalicious: it sounds slightly confusing who the `publisher` here is in first 2 sentences. The person registering the asset? The tool registering the asset? The marketplace? For commons use case, this almost implies as a publisher of data sets I have to run Squid on my local machine.

Something along the lines of maybe: "A publisher or a marketplace publish (register) an asset through Squid. When they do, the local metadata is passed to Squid, which makes some changes and additions in the metadata, puts it into a DDO, and sends that DDO to a metadata store (Aquarius)."

And there should be some link to some Squid document so it becomes clear no real maritime animals are involved in registering an asset

8/README.md
ttmc: While I can understand why `tags` should be an array and not a string, this is what I got from Aquarius at https://aquarius.marketplace.dev-ocean.com/api/v1/aquarius/assets/ddo/did:op:47cd2fa549594bd09e7c024dd5b8252b407fc19cc3f344b8b5b2a4103a4ddc25 

I'm happy to change this but only if the underlying code also gets changed.

8/README.md
ttmc: Okay thanks for spotting that and letting me know. I just removed `found` in commit 21d13e6

8/README.md
ttmc: OEP-11 says that a publisher uses Squid to publish an asset. I guess Squid doesn't have to be running on their local machine though. I will change the wording as you suggest.

I will also add a Squid link.

8/README.md
ttmc: I found some `tags` problems in squid-py and aquarius so I will open some new issues in those repos. Meanwhile I fixed the examples in this pull request.

PR #172, opened 2019-05-31
Support of services in assets metadata

aaitor:


8/README.md
```diff
@@ -183,6 +201,20 @@ The publisher of a DDO MAY add additional attributes or change the above object
         "index": 0
       }
     ],
+    "services": [
```
ttmc: Something like this should also be in the "Remote Metadata" example, but 1) the `url` should be removed (I guess) and 2) there should be an example `encryptedServices` key and value.

8/README.md
```diff
@@ -134,6 +136,22 @@ with the details necessary to consume and validate the data.
...
+
+A service object has the following attributes,
+with the details necessary to consume and validate the data.
+
+| Attribute         | Required | Description                                         |
+| ----------------- | -------- | --------------------------------------------------- |
+| **url**           | (local)  | Service URL. Omitted from the remote metadata. |
+| **index**         | yes      | Index number starting from 0 of the service. |
+| **method**   | yes      | Service method (GET, POST, etc.) |
+| **checksum**      | no       | Checksum of the service using your preferred format (i.e. MD5). Format specified in **checksumType**. If it's not provided can't be validated if the file was not modified after registering. |
```
ttmc: The overall checksum is currently computed using the file checksums (and some other things like author). Should the service checksums also be included when computing the overall checksum? If yes, then that should be mentioned here (in the column describing the overall checksum) and also in OEP-7, which specifies how to compute the overall checksum.

8/README.md
```diff
@@ -107,10 +107,12 @@ Attribute       |   Type        |   Required    | Description
 **price**       | String        | Yes           | Price of the asset in vodka (attoOCEAN). It must be an integer encoded as a string, e.g. "123000000000000000000".
 **files**       | Array of files object | Yes     | Array of File objects including the encrypted file urls. Further metadata about each file is stored: contentType, checksum (optional), content length in bytes (optional), encoding (optional), compression (optional) and remote resourceId (optional)
 **encryptedFiles** | Text         | (remote)    | Encrypted string of the **files** attribute. 
+**services**       | Array of service object | Yes     | Array of Service object including the encrypted service endpoints. Further metadata about each service is stored: method, index, authentication (optional), description (optional), checksum (optional)
```
ttmc: The services Description doesn't need to list all the possible things inside a service object (and whether they are required). It can just reference that section below. The same is true of the Description of the file object, so it could be edited in the same way for consistency. That way, there aren't two places giving the same information, so they can't get out of sync.

PR #177, opened 2019-06-07
Revise 2/COSS regarding using subfolders to expose different versions of an OEP

### LINKED ISSUE #175: Proposal: Put different OEP versions in different subfolders
When an OEP doesn't change a lot (i.e. minor or patch releases), then a new OEP is probably not necessary, but we might want to put the older versions somewhere so they are easy to find. I propose putting them in subfolders of the OEP, like for example:

`8/README.md` with links to versions/files in subfolders, and maybe some comments on status
--- `8/v0.3/README.md`
--- `8/v0.4/README.md`

If we do this, then some part of [2/COSS (OEP-2)](https://github.com/oceanprotocol/OEPs/tree/master/2) should be updated.

ttmc:
Together with pull request #176 (already merged), this pull request resolves #175

### DIFF
--- a/2/README.md
+++ b/2/README.md
@@ -50,7 +50,13 @@ There are three types of OEPs:
 ### OEP Format
 A OEP is a set of Markdown documents (the main file SHOULD be called `README.md`), together with comments, attached files, and other resources. A OEP is identified by its number and short name (e.g. this OEP is **2/COSS**). The number of the OEP is also the name of the directory where its files are stored.
 
-Every OEP (including branches) carries a different number. New versions of the same OEP have new numbers.
+Every OEP (including branches) carries a different number.
+
+If there are different versions of an OEP which don't differ very much, and you want to make them all easy to find, then put them in `README.md` files in separate subfolders, where the subfolder name is the version (e.g. `v0.3`). If using subfolders to expose multiple versions, then the root `README.md` file should have links to the versions/files in the subfolders. Example folder structure:
+
+* `15/README.md`
+* `15/v0.1/README.md`
+* `15/v0.2/README.md`
 
 ### OEP template
 Each OEP MUST customize and include this header:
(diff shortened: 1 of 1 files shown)


kremalicious:
Moving comment over from #175:

I think this introduces too much of a focus on old versions. With subfolders we essentially say: all those versions are on the same level and they could be all in use. And then there're all those new problems being introduced: arbitrary version numbers, defining what justifies a bump, people linking to old versions of an OEP, etc. We should always have one final version, and the rest can be linked to via git commit hashes. I feel replicating an old-school folder structure ignoring what git has to offer is just complicating things, and making updates harder. If people want to go back to an older version they'll use the version history of git.

If we want to link to older versions, we can do so at the end of an OEP, via the commit hash of former version.

I mean, if Ethereum doesn't need different versions of their EIPs, then I'm pretty sure Ocean won't need that either :-) https://github.com/ethereum/EIPs/tree/master/EIPS

ttmc:
I was trying to solve a real problem:

Today, Plecos is checking metadata against the OEP-8 version 0.3 schema, but the live/visible version was 0.4 and Plecos does _not_ support the new things in 0.4 (yet). Moreover, _there is no commit that corresponds to what Plecos checks today_ because there were some errors in the last one labelled 0.3, which I fixed to create the "real" version 0.3, which is what is live now at https://github.com/oceanprotocol/OEPs/tree/master/8/v0.3

The IETF, the home of the RFCs (i.e. the protocol specs for UDP, IP, TCP, FTP, etc.) goes through many versions of specs, and each one is linked. For example (picked randomly), see:

https://tools.ietf.org/html/draft-ietf-ntp-extension-field-07

Screenshot from 2019-06-07 20-57-50

ttmc:
I should probably add that RFCs do eventually end up in a final stable state, and if you want to modify _that_, then you have to make a whole new RFC... which is basically what I was proposing in #175.

trentmc:
We're using a different process to update OEPs: simply update the .md file. Therefore we can close this ticket.

PR #182, opened 2019-07-01
Draft: Data Provenance spec (OEP13)

aaitor:


13/README.md
```diff
@@ -0,0 +1,546 @@
...
+activity. This entity did not exist before generation and becomes 
+available for usage after this generation.
+
+`Generation`, written `wasGeneratedBy(id; e, a, t, attrs)` in PROV-N 
+(a notation for provenance aimed at human consumption), has:
+
+* `id`: an optional identifier for a generation;
+* `entity`: an identifier (e) for a created entity. In Ocean a **DID**;
+* `activity`: an optional identifier (a) for the activity that creates 
+the entity. In Ocean a **DID**;
```
gongf05: both `entity` and `activity` are DID in Ocean. What is the difference between those DID? format or category?

13/README.md
```diff
@@ -0,0 +1,546 @@
...
+* `generatedEntity`: the identifier (e2) of the **derived entity** 
+generated by the derivation. In Ocean a **DID**;
+* `usedEntity`: the identifier (e1) of the **source entity** used by the
+ derivation. In Ocean a **DID**;
+* `activity`: an optional identifier (a) for the activity using and 
+generating the above entities. In Ocean a **DID**;
+* `generation`: an optional identifier (g2) for the generation involving
+ the generated entity (e2) and activity (a). In Ocean a **DID**;
+* `usage`: an optional identifier (u1) for the usage involving the used 
+entity (e1) and activity (a). In Ocean a **DID**;
```
gongf05: Ocean use DID represent so many different entities, activity and etc. Is there a way to distinguish the object of these DIDs? It might be difficult to infer the underlying object by looking at the DID only.

13/README.md
```diff
@@ -0,0 +1,546 @@
...
+...
+
+```
+
+#### Activity
+
+```json
+...
+"activity": {
+    "a1": {
```
gongf05: should activity have a similar DID format such as "did:op:xyz"

13/README.md
```diff
@@ -0,0 +1,546 @@
...
+To guarantee the integrity of the provenance information, the complete 
+`provenance` JSON object inside of the `Provenance` service is 
+serialized and hashed using `Hash.sha3`. 
+
+This generated checksum will be stored as part of the Provenance 
+service in the `checksum` attribute.
+
+Additionally, this checksum can be registered on-chain to prevent the 
+tampering of provenance information. Also this checksum can be used as 
+part of the DID computation.
```
gongf05: Will the provenance information be updated? If any new operation is logged in the provenance information, the checksum needs to be recalculated and on-chain record shall be updated accordingly. An alternative is to create a new provenance record with a new activity, therefore, we keep track of the checksum for new provenance information. 

Another issue is the DID computation using checksum will generate different results as the checksum keeps changing due to new activities.

13/README.md
```diff
@@ -0,0 +1,546 @@
...
+rules:
+
+* All the \n, \t, \r characters must be removed
+* All the whitespaces out of the Json entities names or values must be 
+removed
+* The Json document and all the nested objects must be sort 
+alphabetically
+
+This must generate a JSON object of only one line that can be hashed and
+the verification of the hash can be repeated.
```
gongf05: it is the "minify" or "compact" operation of JSON file :)

13/README.md
aaitor: The difference between DIDs is in the DDO. DDO includes an asset type. In this case an `entity` would be `dataset` and an `activity` a `workflow`

13/README.md
aaitor: DID is only an unique identifier in the network. The details about the `object` represented by this DID are in the DDO.

13/README.md
aaitor: good catch, changing it

13/README.md
aaitor: IMO the data provenance information added to the Asset must not be updated. The intention is to provide an immutable transaction log of how the assets where created and who/how created those assets. Maybe additional provenance information that could be added like `Used` could be tracked out of the Assets DDO.

13/README.md
aaitor: changed :)

aaitor:
@gongf05 , thanks for your comments, much appreciated. I added some clarifications to this thread and amending some of the issues you found.

13/README.md
dgossen: Should we dictate this or not? My feeling is that we should let it transpire naturally as you can make an argument either way (i.e. slowly changing dimensions -> overwrite vs. new record). It shouldn't be up to use to push a specific pattern, but we should support different patterns.

PR #183, opened 2019-07-09
Feature/oep12 simple clarifications

aaitor:


12/README.md
```diff
@@ -215,7 +221,7 @@ Example of a Workflow:
     }
 ```
 
-A Workflow is a new type of Asset (a part of datasets, algorithms, etc.)*[]: 
+A Workflow is a new type of Asset (a part of datasets, algorithms, etc.).
```
jcortejoso: To be an asset means it has to be published on the network and/or it can be purchased? Does the workflow have an attached `did`?

PR #187, opened 2019-07-25
fix typos and minor changes in OEP-12

0x3bfc:
Note:

- The naming convention for a `maxExecutionTime` in workflow is the `wallTime`. 

Questions:
- Will the consumer pay for storing the new generated asset ? Does the locked value include this ?


aaitor:
> Will the consumer pay for storing the new generated asset ? Does the locked value include this ?

Yep, the compute service price includes the storage of the asset

PR #188, opened 2019-08-08
New DDO structure and integrity calculation of DIDs

aaitor:
This is a breaking spec change for Ocean Protocol >= v2.0


12/README.md
```diff
@@ -275,88 +288,195 @@ The complete flow of publishing an asset with a computing service attached is:
...
+							"actorType": "publisher",
+							"handler": {
+								"moduleName": "lockRewardCondition",
+								"functionName": "fulfillAccessSecretStoreCondition",
+								"version": "0.1"
+							}
+						}]
+					},
+					{
+						"name": "serviceExec",
```
H34D: Please don't use abbreviations, we never did it before. Proposal:

ServiceExecution
ServiceExecutionCondition
ServiceExecutionTemplate

aaitor: makes sense, I changed that

PR #195, opened 2019-08-20
update OEP-7 examples, copy editing

kremalicious:
* update all code examples for DDO structure
    * service `index` value from string to integer
    * make all service attribute members consistent: `main`, `additionalInformation`, `curation`
    * valid json, consistent indentation
    * js examples in Ocean js code style
* move change process & language blocks to the end
* update TOC
* add document title at top
* some copy editing and typos

### DIFF
--- a/7/v0.2/ddo-example.json
+++ b/7/v0.2/ddo-example.json
@@ -1,71 +1,85 @@
 {
-	"@context": "https://w3id.org/did/v1",
-	"id": "did:op:e6fda48e8d814d5d9655645aac3c046cc87528dbc1a9449799e579d7b83d1360",
-	"created": "2019-06-26T14:53:09",
-	"updated": "2019-06-30T14:53:09",
-	"publicKey": [{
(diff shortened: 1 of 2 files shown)


7/v0.2/ddo-example.json
```diff
@@ -1,71 +1,85 @@
...
+                "curation": {}
+            }
+        },
+        {
+            "index": 1,
+            "type": "provenance",
+            "serviceEndpoint": "https://service/api/v1/provenance/assets/ddo/did:op:0ebed8226ada17fde24b6bf2b95d27f8f05fcce09139ff5cec31f6d81a7cd2ea",
+            "attributes": {
+                "main": {},
+                "additionalInformation": {}
```
kremalicious: @aaitor I wasn't sure about this change here from `extra` to `additionalInformation` in the provenance service. Or does it need to be `extra`? Cause would be more consistent if this here would be `additionalInformation` just like in the other services

aaitor: It doesn't need to be `extra`. Each individual service could have additional buckets of optional information with names we can't specify upfront.
The only one essential is `main` because will be the one reflecting the information building the service checksum.

PR #196, opened 2019-08-20
Moving serviceAgreementTemplate outside of `main`

aaitor:


11/v0.2/ddo.example.json
```diff
@@ -0,0 +1,246 @@
...
+							"uri": "http://skos.um.es/unescothes/C01194/jsonld",
+							"mediaType": "application/ld+json"
+						},
+						{
+							"uri": "http://skos.um.es/unescothes/C01194/turtle",
+							"mediaType": "text/turtle"
+						}
+					]
+				},
+				"base": {
```
eruizgar91: It should be main here

11/v0.2/ddo.example.json
```diff
@@ -0,0 +1,246 @@
...
+		"owner": "did:op:4c27a254e607cdf91a1206480e7eb8c74856102316c1a462277d4f21c02373b6",
+		"publicKeyBase58": "H3C2AVvLMv6gmMNam3uVAjZpfkcJCwDwnZn6z3wXmqPV"
+	}, {
+		"id": "did:op:b6e2eb5eff1a093ced9826315d5a4ef6c5b5c8bd3c49890ee284231d7e1d0aaa#keys-3",
+		"type": "RsaPublicKeyExchangeKey2018",
+		"owner": "did:op:5f6b885202ffb9643874be529302eb00d55e226959f1fbacaeda592c5b5c9484",
+		"publicKeyPem": "-----BEGIN PUBLIC KEY...END PUBLIC KEY-----\r\n"
+	}],
+	"service": [{
+			"metadata": {
```
eruizgar91: This should be attributes

11/v0.2/ddo.example.json
```diff
@@ -0,0 +1,246 @@
...
+				},
+				"curation": {
+					"numVotes": 123,
+					"rating": 0.0,
+					"schema": "Binary Votting"
+				}
+			},
+			"serviceDefinitionId": "0",
+			"serviceEndpoint": "http://localhost:5000/api/v1/aquarius/assets/ddo/{did}",
+			"type": "Metadata"
```
eruizgar91: This should be lowercase

ssallam: Type is actually `Metadata` in current squid-py implementation.

aaitor: yep, but in the new versions of the specs all the service types are in lower case to avoid discrepancies.

PR #215, opened 2019-11-18
OEP-18: Data Token

0x3bfc:
@gongf05 :point_down:


0x3bfc:
I don't see that we need to have a new contract for each DID registered on chain. 

```Data tokens should be mint as you go```

According to the [ERC-721](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-721.md) specification and the contract interfaces we have the following:

```javascript
function safeTransferFrom(address _from, address _to, uint256 _tokenId, bytes data) external payable;
```
This means:
- A DID owner can mint any ***N*** amount of data tokens
- Each data token has incremented & unique `_tokenId`
- `_tokenId` is incremented by the smart contract prior the mint
- A DID  is stored in the `bytes data`
- No one can mint new data token except DID owner.
- A Data tokens is ***transferrable***
- A Data token is ***burnable*** (in the context of the SEA)
- A Data token can be consumed only once

0x3bfc:
By using data tokens in the context of SEA, It does not need new conditions. However It will only require a simple contract upgrade by overriding the fulfilment method in [AccessSecretStoreCondition](https://github.com/oceanprotocol/keeper-contracts/blob/develop/contracts/conditions/AccessSecretStoreCondition.sol) contract to accept the data token, process its validity, set `permissions` to true for this `grantee`, and finally burn the data token. 
```javascript
function fulfill(
        bytes32 _agreementId,
        bytes32 _documentId, // DID
        address _grantee,
        uint256 _tokenId
    )
        public
        returns (ConditionStoreLibrary.ConditionState)
```
On the other hand,  while consuming the asset, secret store automatically calls `checkPermissions` (Built-in functionality in secret store) and re-encrypts the URL using the consumer's public key:
```javascript
function checkPermissions(
        address _grantee,
        bytes32 _documentId
    )
        external view
        returns(bool permissionGranted)
```

0x3bfc:
On the level of template:

We will use the same `EscrowAccessSecretStoreTemplate` because we just use the same conditions for the traditional data access use case


:nerd_face:

trentmc:
Done

PR #220, opened 2020-01-28
OEP-12: compute updates [WIP]

0x3bfc:
This PR fixes the existing definition of the compute service offered by Ocean Protocol.


kremalicious:
change request: move `attributes.main.serviceAgreementTemplate` to `attributes.serviceAgreementTemplate`, just like in `access` service.

Use case: I created a universal method over in squid-js to order any service and there I have to grab the template name from the service and it would be preferable to have this always in the same place:
https://github.com/oceanprotocol/squid-js/pull/350/files#diff-6ee8f9b5b76b7bfbd2b9daa6a59bab57R38

12/README.md
```diff
@@ -3,47 +3,46 @@ shortname: 12/EXEC
...
+    		* [Volumes](#volumes)
+ 			* [Network isolation](#network-isolation)
 
 
 
 ---
 
 
-# Execution of Compute Services using Service Agreements
+# Introduction
```
kremalicious: there should be only one h1. For the structure I suggest to follow how we write all our last OEPs, e.g. take a look at OEP-8 https://github.com/oceanprotocol/OEPs/blob/master/8/v0.4/README.md

PR #233, opened 2021-04-14
add isDisabled

alexcos20:


7/v0.3/README.md
```diff
@@ -102,6 +102,8 @@ A DDO document is composed of standard DDO attributes:
 * `verifiableCredential`
 * `dataToken`
 * `service`
+* `isDisabled`   - optional flag, if set, will disable asset consumption, but the asset will appear in searches & teasers
```
mihaisc, line 105: remove `& teasers`

7/v0.3/README.md
```diff
@@ -102,6 +102,8 @@ A DDO document is composed of standard DDO attributes:
...
+* `isRetired`   - optional flag, if set, will disable the display of the asset & consumption
```
mihaisc, line 106: not sure if we need to specify here `but the asset will appear in searches` or `will disable the display of the asset` since this is just our decision in the market. Other markets can handle this differently

alexcos20: we need to have a description for the flags and most importantly why isDisabled != isRetired

mihaisc: ok then

mihaisc: i would suggest to add to `isDisabled` something that it's scope is temporary. Meaning you should use this if you want to disable the asset for a short period of time, whilst `isRetired` is more or less permanent

alexcos20: updated

kremalicious: not sure how the retire got sneaked in here but we need to discuss its intention and flow across the stack. So far retiring in our stack means we delete the metadata, at least that's what the code tells me. How is this supposed to change? What is the implication of that flag across our stack? Is only market going to use it?

mihaisc: well, i don't really now why curation is still here. This `isListed` is similar to `isInPurgatory`

alexcos20: agreed, removed

mihaisc (approved):

PR #236, opened 2021-05-07
Status flags updates

kremalicious:
To give all the `is..` things a consistent place I'm proposing to actually use the `curation` attribute within the metadata service for everything which is signaled by the publisher.

Because it is actual curation, the publisher curates the status of their data set on their own, all of which should only happen within the metadata service. All other root DDO things like `isConsumable` or `isInPurgatory` are not set by the publisher but added automatically so it quickly can become confusing to know which attributes are updated by publisher and which ones by Aquarius itself. Sticking to the principle that all metadata the publisher updates is within the metadata service keeps it more transparent and consistent how those values can change.

So we would get this as the default within the meta data service:

```json
  "curation": {
    "isListed": true,
    "isRetired": false,
    "isOrderDisabled": false
  }
```

---

Requires updates in:

- [x] ocean.js (https://github.com/oceanprotocol/ocean.js/pull/789)
- [x] ocean.py
- [ ] plecos?
- [ ] aquarius?

---

Implications for:

- https://github.com/oceanprotocol/ocean.js/pull/697
- https://github.com/oceanprotocol/market/issues/452
- https://github.com/oceanprotocol/market/pull/463
- https://github.com/oceanprotocol/market/issues/445
- https://github.com/oceanprotocol/ocean.py/pull/314
-

### DIFF
--- a/7/v0.3/README.md
+++ b/7/v0.3/README.md
@@ -103,16 +103,14 @@ A DDO document is composed of standard DDO attributes:
 * `verifiableCredential`
 * `dataToken`
 * `service`
-* `isDisabled`   - optional flag, if set, will disable asset consumption, but the asset will appear in searches. This a temporary flag, publisher can switch it any time.
 * `credentials`   - optional flag, which describes the credentials needed to access a dataset (see below)
 
-
-
-Asset metadata can be included as one of the objects inside the `"service"` array, with type `"metadata"`.
+Asset metadata must be included as one of the objects inside the `"service"` array, with type `"metadata"`.
 
 #### DDO Services
 
 Each type of asset (dataset, algorithm, workflow, etc, ..) typically will have associated different kind of services. There are multiple type of services that are commonly added to all the assets:
+
 * metadata - describing the asset
 * provenance - describing the asset provenance
 * access - describing how the asset can be downloaded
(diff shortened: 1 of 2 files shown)


mihaisc:
`isConsumable` is already deprecated and will soon be removed.

mihaisc:
But wait, why is this in curation?

jamiehewitt15:
It would be good to get a conclusion on this, the Daimler team are waiting for it in order to complete: https://github.com/oceanprotocol/ocean.js/pull/786.

@kremalicious Also, just to clarify, would any of these fields be required or have a default value?

alexcos20:
after a call, we agreed on the following:

```json
  "status": {
    "isListed": true,
    "isRetired": false,
    "isOrderDisabled": false
  }
```

calina-c:
> after a call, we agreed on the following:
> 
> ```json
>   "status": {
>     "isListed": true,
>     "isRetired": false,
>     "isOrderDisabled": false
>   }
> ```

Can you explain the logic here, a bit? Previously we had simply "isDisabled". Should I take all of these into account now when checking if the asset is consumable? And, if so, should there be different response codes?

jamiehewitt15:
Also, can I double check if any of them are required or have a default value?

